PR #5, opened 2015-10-14
Sending empty arrays or dictionaries may raise an exception

fejesd:
eg. **emit('test message',{})** will fail due to functools.reduce doesn't support empty sequences. This fix seems to work for me flawlessly, maybe you will find a more elegant solution.


miguelgrinberg:
Thanks. I have addressed this by adding an initializer to the reduce function.

PR #22, opened 2016-03-08
Avoid KeyError in is_connected.

chrisdoehring:
Change in is_connected to avoid a KeyError.


miguelgrinberg:
Looks good. Thanks!

PR #30, opened 2016-06-26
Django example

sillygod:
I write a simple usage example for django.


sillygod:
I will make another pull request because original python-socketio has updated that cause conflict...

Issue #24, opened 2016-03-28
Working with sockets from wsgi app


Qwasser:
Hello!
I am trying to work with sockets from django app. I don't understand how can I possibly access to sockets from django. For example, I have request in django that creates some obkects and I want tu brodcast this object to all of the clients. Could you give a hint, please?

miguelgrinberg:
Unfortunately there is no integration with Django at this point, so you will need to write the glue code that connects your Django app with the SocketIO server. The example in this repository shows how you can integrate SocketIO with a Flask server. This should translate fairly easily into Django. If you do make a Django based example app I would love to include it here as an example, by the way.

Another alternative would be to create the SocketIO server separately from your Django app, and then have the Django app connect to it via a message queue to pass on updates.

sillygod:
@Qwasser 

Hello ~

I've written a example with django and make a pull request. Maybe you can see the [pull request here](https://github.com/miguelgrinberg/python-socketio/pull/30)

Regards,
Jing

eddyekofo94:
Integration with a Pyramid app, any examples yet?
Thank you.

miguelgrinberg:
Not yet. I would happily accept one as a contribution if you end up doing it yourself.

quietcoolwu:
@miguelgrinberg Do you mean the example used by js code in `index.html`?

miguelgrinberg:
@quietcoolwu see the example for Django: https://github.com/miguelgrinberg/python-socketio/tree/master/examples/server/wsgi/django_example

quietcoolwu:
@miguelgrinberg thanks for your quick response!
yeah, the `javascript` code in `index.html` in this dir is a must-need now? If we removed the html, will python-socketio embedded in `django` still work?
I found that you claimed `Unfortunately there is no integration with Django at this point, so you will need to write the glue code that connects your Django app with the SocketIO server`, I wonder do we still need to hack the django now?

miguelgrinberg:
The JavaScript code is the client side application, it has nothing to do
with Django. You can remove that and provide your own client in a separate
project if you prefer.

PR #31, opened 2016-06-28
add django example for python-socketio

sillygod:
I've resolved the conflict and make a pull request again. #24


miguelgrinberg:
Oh, this is nice, awesome to have a Django example here. I'll look at this example in the next couple of days and let you know if I need you to make any changes. Thanks!

sillygod:
@miguelgrinberg 

Thanks for this great implement socketio server with python. In fact, this example is my experimental project. I just find some trick or way to interact with django framework and I found a bug or problem in developing this example. 

That is, I can not send **message in Chinese** with `gevent` wsgi server but `eventlet` server works so in my example I use eventlet.

miguelgrinberg:
Why did you write your example for Django 1.8 and not 1.9?

sillygod:
@miguelgrinberg 

Q1:

There are many apis changed from version 1.8 to version 1.9 and not backward compatible more detail you see [here](https://docs.djangoproject.com/en/1.9/releases/1.9/). Actually, my company project still use django with version 1.8 because many third party library does not support django 1.9 although some top library may support (ex. rest_framework). That's why I choose django 1.8. For me, it is something like python 2 and python 3. lol.

Q2:

There is just small changed in my copy version `SocketMiddleware`. 

```
query = urllib.parse.parse_qs(environ.get('QUERY_STRING', ''))
sid = query.get('sid', None)

if sid is None:
    self.wsgi_app(environ, start_response)

```

I can not find any way to get framework specific instance (ex. django's request) so I make an entrance for that. see the following code snippet

```
@csrf_exempt
def socketio(request):
    try:
        request.environ['django_request'] = request
    except:
        logging.getLogger("socketio").error("Exception while handling socketio connection", exc_info=True)
    return HttpResponse(200)

urls = patterns("", (r'', socketio))
```

We must enter django's request-response cycle to get the `request` and set it to `environ`. 

```
query = urllib.parse.parse_qs(environ.get('QUERY_STRING', ''))
sid = query.get('sid', None)

if sid is None:
    self.wsgi_app(environ, start_response)

```
So first time, the request enter the middleware. it will not has sid in its query so let it enter the django request-response cycle.

To sum up,
I just find a trick to get framework specific instance.  Is there any way to get the same effect ? I would appreciate it if you tell me :)

Q3:

About that,  I will add it to the example. I forgot it.

Q4:

The reason I use class based style is I want to split socket related code and server code. In your example, you must define socket function in the file where server is defined. Furthermore, in my opinion, class based style is more easier to manage project's code. 

about this problem `The problem is that there are other decorators for error handling that don't appear to be supported in this model`. Sorry for my english. I can not fully understand your description.  Could you please give me a concrete example.

examples/django_chat/chat/management/commands/runserver_socketio.py
```diff
@@ -0,0 +1,108 @@
...
+from time import sleep
+from os import getpid, kill, environ
+from signal import SIGINT
+import six
+import copy
+
+from django.conf import settings
+from django.core.handlers.wsgi import WSGIHandler
+from django.core.management.base import BaseCommand, CommandError
+from django.core.management.commands.runserver import naiveip_re, DEFAULT_PORT
```
miguelgrinberg: I believe the only issue when running your example under Django 1.9 is that `DEFAULT_PORT` isn't defined. At least that's what I see here when I test.

examples/django_chat/chat/sockets.py
```diff
@@ -0,0 +1,64 @@
...
+
+@namespace('/test')
+class TestNamespace:
+
+    def __init__(self, name):
+        self.name = name
+        self.request = None  # django request object
+
+    def _get_socket(self, sid):
+        socket = namespace.server.eio._get_socket(sid)
```
miguelgrinberg: The `eio._get_socket()` method is private. The `_` prefix indicates it is not meant to be called from the outside. Is there any way you can do what you need to do without using the internal socket object? I really don't like the idea of promoting its use in an example application.

examples/django_chat/chat/sockets.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def on_disconnect_request(self, sid):
+        namespace.server.disconnect(sid, namespace='/test')
+
+    # two method must have
+    def on_connect(self, sid, environ):
+        if 'django_request' in environ:
+            request = environ['django_request']
+            socket = self._get_socket(sid)
+            socket._request = request
```
miguelgrinberg: It seems you are putting the Django request object as an extra member in the internal socket object. It is actually bad practice to inject new members into objects you get from another package. In addition to that, you are doing this on the connection event only, so in the case of long-polling, where new requests are issued all the time, this `_request` object that you saved would be stale.

examples/django_chat/sdjango/sd_middleware.py
```diff
@@ -0,0 +1,64 @@
...
+
+        path = environ['PATH_INFO']
+        if path is not None and \
+                path.startswith('/{0}/'.format(self.engineio_path)):
+
+            query = urllib.parse.parse_qs(environ.get('QUERY_STRING', ''))
+            sid = query.get('sid', None)
+
+            if sid is None:
+                self.wsgi_app(environ, start_response)
```
miguelgrinberg: I think I understand what you are trying to do here, but I don't think this is a good solution. As I mentioned above, there isn't a single request, during the Socket.IO communication the client may send lots of requests, and here you are only saving the first. Not familiar with Django internals, but isn't there a way to create a request object from the `environ` dict?

examples/django_chat/sdjango/sd_manager.py
```diff
@@ -0,0 +1,22 @@
+from socketio.base_manager import BaseManager
+
+
+class SdManager(BaseManager):
```
miguelgrinberg: The `BaseManager` class is not supposed to be used from the outside. Besides, this subclass that you created doesn't do anything, so why do you have it?

examples/django_chat/chat/urls.py
```diff
@@ -0,0 +1,13 @@
...
+    url, patterns, include
+)
+
+import sdjango
+
+from .views import socket_base
+
+
+urlpatterns = [
+    url(r'^socket\.io', include(sdjango.urls)),
```
miguelgrinberg: This defeats the purpose of using a middleware to do the routing between the socket and django endpoints. Note that in the middleware the `/socket.io` endpoint can be configured to be anything you want, while here it is hardcoded, so the ability to rename the endpoint is broken. I think I prefer a soluiton that does not require hacks around the middleware. I will think about how to get the request object using a cleaner way.

miguelgrinberg:
> The reason I use class based style is I want to split socket related code and server code. In your example, you must define socket function in the file where server is defined.

You do not have to have the socket and http endpoints together, I have written larger applications that have them in separate modules or packages, while still using the decorator syntax for socket events.

>  Furthermore, in my opinion, class based style is more easier to manage project's code.

Well, yes, this is totally a matter of preference. The problem is that if you change the style to class-based, then a portion of this project's documentation is invalidated, and now you are asking people to follow a structure that is not documented.

examples/django_chat/sdjango/sd_manager.py
sillygod: yes, this can be removed. Originally, I plan to do some glue code for django here. However, it seems that nothing can do here.

examples/django_chat/sdjango/sd_middleware.py
sillygod: In my understanding, it is true that the same client can send lots of requests but they are almost the same aren't they? That why I only save it once. 

I have no idea how to create request object from the environ.  I think it is no way. 
`request` instance is produced in django request-response cycle so we must enter wsgi app (django) to get it.

examples/django_chat/chat/sockets.py
sillygod: wow,  so in the case of long-polling, the `on_connect` function will be called multi times??

examples/django_chat/chat/sockets.py
sillygod: yes, I undertstand `_` prefix's meaning in python. I think every client has its own socket so I want the django request instance to bind to it

examples/django_chat/chat/sockets.py
miguelgrinberg: No, the connection event will happen only the first time. But in later requests, you will still be giving event handlers access to the connection request. It seems the correct thing to do would be either to provide access to the correct request object, or else not offer access to a request, since really this is a different path and the concept of an HTTP request makes really no sense.

miguelgrinberg:
I have added a Django application that is a simplified version of this one, so I'm closing this PR.

quietcoolwu:
@miguelgrinberg Hi Miguel, do we have django support now?(1.11 under python 2.7)

miguelgrinberg:
@quietcoolwu Yes, and there is a Django example in this repository.

PR #41, opened 2016-08-16
Add wildcard event

logarytm:
So I found myself in need to capture any incoming event. This patch introduces the wildcard (notice no plural form) event, which is basically triggered for every message coming from the client.

The API is pretty similar except the event name is passed as the first argument:

``` python
@sio.on('*')
def catch_em_all(event, *data):
    ...
```

### DIFF
--- a/socketio/server.py
+++ b/socketio/server.py
@@ -397,6 +397,7 @@ class Server(object):
         namespace = namespace or '/'
         self.logger.info('received event "%s" from %s [%s]', data[0], sid,
                          namespace)
+        self._trigger_event('*', namespace, sid, *data)
         r = self._trigger_event(data[0], namespace, sid, *data[1:])
         if id is not None:
             # send ACK packet with the response returned by the handler
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -269,6 +269,13 @@ class TestServer(unittest.TestCase):
         s._handle_eio_message('123', '2["my message","a","b","c"]')
         handler.assert_called_once_with('123', 'a', 'b', 'c')
 
+    def test_handle_event_wildcard(self, eio):
+        s = server.Server()
+        handler = mock.MagicMock()
+        s.on('*', handler)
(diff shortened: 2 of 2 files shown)


miguelgrinberg:
I was planning to implement a similar feature, but to be consistent with Flask, I thought about it as a `before_event` handler. See https://github.com/miguelgrinberg/Flask-SocketIO/issues/265.

I think seeing it as a before event makes some corner cases easier to think about, since these are issues that Flask has resolved nicely. For example, having the catch all somehow prevent the actual event handler from running under certain conditions can be useful (such as for authorization checks).

That said, I really like the simplicity of your solution, Sounds like a good base to implement my `before_event` idea.

gordol:
I've been tinkering around with a similar idea, but I'd like to avoid always firing an additional event, unless there are listeners... Is that how this works, or does it always fire the '*' event, regardless of specific wildcard handlers being present? Am I overcomplicating this? I do like the simplicity of this patch, with a single line of code to implement the feature, but I feel like it could potentially introduce a good bit of overhead on busy servers. Any thoughts?

WildBamaBoy:
Was something like this ever implemented? I have events with variable names like `upload_d7h345lf' that I cannot catch, and this would be great to have as I cannot _easily_ change the client-side that's creating these events.

miguelgrinberg:
@WildBamaBoy You can implement a catch-all event handler using class-based namespaces. You just need to override the [trigger_event](https://github.com/miguelgrinberg/python-socketio/blob/master/socketio/namespace.py#L23) method in your namespace class and dispatch the event according to your own rules.

gordol:
@WildBamaBoy you could use a middleware to rewrite the event names as they come in... if you don't want to alter socket.io code.

Also, hello from a fellow Alabamian. ;)

gordol:
@miguelgrinberg 's solution is better though. ;)

miguelgrinberg:
Wildcard event handlers can be implemented using class-based namespaces, so I'm going to close this.

PR #44, opened 2016-08-21
Optionally pass current environ to event handlers

efficiosoft:
Pendant of https://github.com/miguelgrinberg/python-engineio/pull/27.

It basically works the same way.

### DIFF
--- a/socketio/server.py
+++ b/socketio/server.py
@@ -56,12 +56,16 @@ class Server(object):
                                  default.
     :param cors_credentials: Whether credentials (cookies, authentication) are
                              allowed in requests to this server.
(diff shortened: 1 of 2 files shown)


miguelgrinberg:
As I mentioned in https://github.com/miguelgrinberg/python-engineio/pull/27, I really see no compelling reason to expose the WSGI environment to every event handler. And even if there was a good reason, there isn't a one-to-one relationship between WSGI requests and events, sometimes the server has two outstanding WSGI requests from the client (long-polling) and sometimes it has none (WebSocket).

efficiosoft:
I'd suggest discussing in https://github.com/miguelgrinberg/python-engineio/pull/27 to have one single thread to follow.

PR #45, opened 2016-08-23
Event handler interceptors

efficiosoft:
This deals with the concept of event handler middlewares. I wrote a chapter in the docs about them. Basically it's a way to add before/after event handlers on three layers for fine-grained control: These layers can alter the arguments the event handler will get or its return value transparently or completely abort execution, returning something else.
- at server layer for _every_ event by adding the middleware to `socketio.Server.middlewares`
- at namespace layer by adding the middleware to `socketio.Namespace.middlewares`
- at the event handler layer by decorating an event handler with `@util.apply_middleware(...)`

Finally, when the event is triggered, the `socketio.Server._trigger_event` method collects middlewares for the used event handler from these three layers (from least to most specific one) and applies them on the handler before execution.

I wrote a test for it that is commented and it works well.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -47,6 +47,7 @@ client:
 
     import socketio
     import eventlet
+    import eventlet.wsgi
     from flask import Flask, render_template
 
     sio = socketio.Server()
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -213,6 +213,37 @@ methods in the :class:`socketio.Server` class.
 When the ``namespace`` argument is omitted, set to ``None`` or to ``'/'``, the
 default namespace, representing the physical connection, is used.
 
+Class-Based Namespaces
+----------------------
+
+As an alternative to the decorator-based event handlers, the event handlers
+that belong to a namespace can be created as methods of a subclass of 
+:class:`socketio.Namespace`::
+
+    class MyCustomNamespace(socketio.Namespace):
+        def on_connect(sid, environ):
+            pass
+
+        def on_disconnect(sid):
+            pass
+
+        def on_my_event(sid, data):
+            self.emit('my_response', data)
+
+    sio.register_namespace(MyCustomNamespace('/test'))
+
+When class-based namespaces are used, any events received by the server are
+dispatched to a method named as the event name with the ``on_`` prefix. For
+example, event ``my_event`` will be handled by a method named ``on_my_event``.
+If an event is received for which there is no corresponding method defined in
+the namespace class, then the event is ignored. All event names used in
+class-based namespaces must used characters that are legal in method names.
+
+As a convenience to methods defined in a class-based namespace, the namespace
+instance includes versions of several of the methods in the 
+:class:`socketio.Server` class that default to the proper namespace when the
+``namespace`` argument is not given.
+
 Using a Message Queue
 ---------------------
 
@@ -456,6 +487,8 @@ API Reference
 .. autoclass:: Middleware
    :members:
 .. autoclass:: Server
+   :members:
+.. autoclass:: Namespace
    :members:
 .. autoclass:: BaseManager
    :members:
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -265,18 +265,35 @@ subset of them:
 * ``after_event(*args)`` is called after the event handler with the
   event name, the namespace and the list of values the event handler
   returned. It may alter that values eventually.
-* ``ignore_event`` is called before the middleware is applied to an
+* ``handle_exception`` is called when an exception is raised during the
+  event handling process with the event name, namespace and the exception
+  object as its arguments. The exception could have been raised by either
+  the event handler itself, or the ``before_event`` and ``after_event``
+  methods of another middleware that comes later in the middleware chain
+  (one that is applied more-specifically).
+* ``ignore_for`` is called before the middleware is applied to an
   event handler with the event name and namespace as arguments. If its
-  return value resolves to ``True`` the middleware is not applied to that
-  particular event handler.
-
-If one of these methods returns something else than ``None``, execution
-is stopped at that point and the returned value is treated as if it was
-returned by the event handler.
+  return value resolves to ``True`` the middleware is ignored and not
+  applied to that particular event handler.
 
 Event handler middlewares can be chained. The ``before_event`` methods
 will be executed in the same order the middlewares were added, the
 ``after_event`` methods are called in reversed order.
+
+If the ``before_event`` method returns something else than ``None``,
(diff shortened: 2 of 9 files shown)


efficiosoft:
I hope that I have not muddled up the documentation after this merging. But I needed to bring it in sync with master.

efficiosoft:
Here is an example of how I use it in real world. This packs and unpacks all arguments for an event handler with ``msgpack`` transparently.

```
import msgpack

class CodingMiddleware:
  def ignore_event(self, event, namespace):
    return event in ("connect", "disconnect")

  def before_event(self, event, namespace, args):
    assert len(args) == 2  # sid and packed args
    arg = None
    assert isinstance(args[-1], bytes)
    unpacked_args = msgpack.loads(args[-1], encoding="utf-8")
    assert isinstance(unpacked_args, list)
    args.pop()
    args.extend(unpacked_args)

  def after_event(self, event, namespace, ret):
    packed_ret = msgpack.dumps(ret, use_bin_type=True)
    ret.clear()
    ret.append(packed_ret)
```

I simply add it to all event handlers of all namespaces:
```
sio = socketio.Server()
sio.middlewares.append(CodingMiddleware())
```

efficiosoft:
Ah, you could additionally implement a ``@sio.before_event`` decorator that creates a middleware under the hood and adds it to the server, that way you won't loose the easy interface you talked about while keeping middlewares as powerful as they are in my implementation.

Something like that:
```
import types

class Server:
[...]
    def before_event(self, before_handler):
        class Middleware:
            pass

        mw = Middleware()
        mw.before_event = types.MethodType(before_handler, mw)
        self.middlewares.append(mw)
        return before_handler
[...]
```

...and the same for ``@sio.after_event``.

efficiosoft:
If I look at my code now, I'd probably rename the ``ignore_event`` method of the middleware object to ``ignore_for`` to express what it does more clearly.

efficiosoft:
I think I found another way to describe the whole idea with my humble English skills.

Summary of interceptors
=======================

Interceptors are the idea of splitting the event processing horizontally into slices where each slice is an interceptor.

One interceptor may only depend on other interceptors that are placed above it (above means added later).

Event processing is done as follows:
- Arguments of an event handler are passed into each interceptors ``before_event`` method from bottom to top. Each of these interceptors may alter the arguments.
- When the top is reached, the real event handler is called with the updated arguments.
- The returned values are now passed to the topmost interceptor's ``after_event`` method and may be altered there. The arguments ripple down until the bottommost interceptor is reached. Afterwards, the updated value is used as the result of the event processing.

There are special cases that interrupt the process and continue it at another point:

- When a ``before_event`` method returns something else than ``None``, the "climbing" of the interceptor hill is aborted and the ``after_event`` processing starts at the interceptor that caused the abortion, using the returned values of ``before_event``. The real event handler is never called in that case.
- When an exception is raised *anywhere* in the whole process (including ``before_event`` and ``after_event`` calls as well as the real event handler), a ``handle_exception`` method is called with the raised exception as its argument in the interceptor we've currently climbed up to. If an exception occurs in the real event handler, this will be the topmost interceptor. From there on, we start to climb downwards. The method can:
  - either handle the exception and return something that will be treated as the return value, which is then passed to the next lower ``after_event`` method
  - or re-raise the exception, what causes the next lower interceptor to be considered to handle the exception.

I hope this makes the procedure clearer. It might seem to be a bit complex at first, but it actually is simple to use.

efficiosoft:
I just merged your master branch into this and resolved a minor conflict to make it compatible with master again. What do you think about it?

This also addresses https://github.com/miguelgrinberg/python-socketio/pull/41.

AndreiD:
bump ?

miguelgrinberg:
I'm not ignoring this, I've just been thinking about this feature and while it does have some interesting ideas I'm still not sure about it.

First of all, it seems most of what this extension does can be easily implemented by the application inside a class-based namespace. Something like this:

    class ClassBasedInterceptor(Namespace):
        def ignore_for(self, event, *args):
            return False

        def before_event(self, event, *args):
            return args

        def after_event(self, event, ret):
            return ret

        def exception_handler(self, exc):
            raise exc

        def trigger_event(self, event, *args):
            try:
                if not self.ignore_for(event, *args):
                    new_args = self.before_event(event, *args)
                    ret = super(MyNamespace, self).trigger_event(event, *new_args)
                    return self.after_event(event, ret)
            except Exception as exc:
                return self.exception_handler(exc)

Isn't the above namespace almost equivalent to the functionality of your patch?

I think the only thing this does not have is the chaining of multiple interceptors. That is actually one of the problems I have with this feature. WSGI middlewares are very friendly to chaining because they can run pretty much in any order. Each middleware takes two arguments, and it passes them on to the next middleware, either modified or not, but there are always the same two arguments. Your interceptor design can modify the list of arguments, and this makes the order in which interceptors run very important. If the interceptor somehow get added in the wrong order, then things are not going to work.

efficiosoft:
Basically, the code you posted could be used as an interceptor, but you won't be able to add interceptors to single handlers only. my ``apply_interceptor`` and ``ignore_interceptor`` decorators allow that.

Second, there seems to be a missunderstanding on how arguments are passed to ``before_event``. They are not passed with ``*args``. Instead, a list with those arguments is passed that, of course, is mutable. So, ``before_event`` is always called with four arguments: ``self, event, namespace, args``.

miguelgrinberg:
Say you have `intercept1` and `intercept2` interceptors. Let's say intercept1's before_event takes all the arguments and concatenates them into a single string, separated by commas. Intercept2's before_event takes a single argument, and breaks it into multiple arguments separating by commas.

So for example, client invokes the event with the following arguments:

    'abc,def', 'ghi'

Do you see how the order of the interceptors will change the outcome?

If intercept1 runs before intercept2, then you end up with three arguments sent to the actual event:

    'abc', 'def', 'ghi'

But if intercept2 runs before intercept1, then there is a single final argument:

    'abc,def'

Note how in this second case, since intercept2 expects one argument but gets two, it has to throw away the second argument.

This is what I mean when I say that the interceptors as you have defined do not lend to chaining unless the order is enforced.

efficiosoft:
Of course, if arguments get altered by interceptors, you can't do that without enforcing a specific order. That's clear. It would be the same trying to encode with gzip and then encrypt by AES and decode/decrypt it the other way around.

miguelgrinberg:
This issue will be automatically closed due to being inactive for more than six months. Seeing that I haven't responded to your last comment, it is quite possible that I have dropped the ball on this issue and I apologize about that. If that is the case, do not take the closing of the issue personally as it is an automated process doing it, just reopen it and I'll get back to you.

Issue #61, opened 2016-12-21
Are SIDs shared between namespaces? 


AndreiPashkin:
Are SIDs shared between namespaces?  Is it possible to handle events from all namespaces in a handler for the default namespace?

miguelgrinberg:
> Are SIDs shared between namespaces?

No, each time the client calls the `connect` function a new SID is generated.

> Is it possible to handle events from all namespaces in a handler for the default namespace?

There is no wildcard to handle all namespaces, but you can decorate a single handler function with multiple `@on()` decorators for multiple namespaces.

miguelgrinberg:
This issue will be automatically closed due to being inactive for more than six months. Please reopen if you need more assistance.

PR #62, opened 2016-12-21
Clarify level of separation of namespaces

AndrewPashkin:
1. Add mentioning, that each namespace has own SID's. 
   To make it clear for a reader, that they are not global.
2. Remove the phrase, that the default namespace represents "physical connection".
    - [Socket.IO docs][1] does not mention anything like that.
    - It makes user think, that the default namespace is somehow special and maybe a "catch-all" namespace.
    - What is "physical connection" and how it represents is? Socket.IO is a pretty high level protocol, there is nothing in it that represents anything "physical".

[1]: http://socket.io/docs/rooms-and-namespaces/#

It is related to #61.


miguelgrinberg:
Thanks, I merged with just a tiny correction.

PR #64, opened 2017-01-04
add zmq manager for zmq as message queue transport

etseidler:
this requires a corresponding change to flask-socketio, for which I have filed a separate pull request.

### DIFF
--- a/socketio/zmq_manager.py
+++ b/socketio/zmq_manager.py
@@ -102,7 +102,10 @@ class ZmqManager(PubSubManager):  # pragma: no cover
     def _listen(self):
         for message in self.zmq_listen():
             if isinstance(message, str):
-                message = pickle.loads(message)
+                try:
+                    message = pickle.loads(message)
+                except Exception:
+                    pass
             if isinstance(message, dict) and \
                     message['type'] == 'message' and \
                     message['channel'] == self.channel and \
--- a/socketio/zmq_manager.py
+++ b/socketio/zmq_manager.py
@@ -50,7 +50,7 @@ class ZmqManager(PubSubManager):  # pragma: no cover
     """
     name = 'zmq'
 
-    def __init__(self, url='tcp://localhost:5555+5556',
+    def __init__(self, url='zmq+tcp://localhost:5555+5556',
                  channel='socketio',
                  write_only=False):
         if zmq is None:
--- a/socketio/zmq_manager.py
+++ b/socketio/zmq_manager.py
@@ -1,0 +1,110 @@
+import time
+import pickle
+import re
+
+try:
+    import zmq
+except ImportError:
+    zmq = None
+
+from .pubsub_manager import PubSubManager
+
+
+class ZmqManager(PubSubManager):  # pragma: no cover
+    """zmq based client manager.
+
+    This class implements a zmq backend for event sharing across multiple
+    processes. To use a zmq backend, initialize the :class:`Server` instance as
+    follows::
+
+        url = 'zmq+tcp://hostname:port1+port2'
+        server = socketio.Server(client_manager=socketio.ZmqManager(url))
+
+    :param url: The connection URL for the zmq message broker,
+                which will need to be provided and running.
+    :param channel: The channel name on which the server sends and receives
+                    notifications. Must be the same in all the servers.
+    :param write_only: If set to ``True``, only initialize to emit events. The
+                       default of ``False`` initializes the class for emitting
+                       and receiving.
+
+
+    N.B.
+    a zmq message broker must be running for the zmq_manager to work.
+    you can write your own or adapt one from the following simple broker below.
+    port numbers in the broker must match port numbers in connection string.
+    ``
+    import zmq
+
+    receiver = zmq.Context().socket(zmq.PULL)
+    receiver.bind("tcp://*:5555")
+
+    publisher = zmq.Context().socket(zmq.PUB)
+    publisher.bind("tcp://*:5556")
+
+
+    while True:
+        publisher.send(receiver.recv())
+    ``
+
+    """
+    name = 'zmq'
+
+    def __init__(self, url='tcp://localhost:5555+5556',
+                 channel='socketio',
+                 write_only=False):
+        if zmq is None:
+            raise RuntimeError('zmq package is not installed '
+                               '(Run "pip install pyzmq" in your '
+                               'virtualenv).')
+
+        r = re.compile(':\d+\+\d+$')
+        if not (url.startswith('zmq+tcp://') and r.search(url)):
+            raise RuntimeError('unexpected connection string: ' + url)
+
+        url = url.replace('zmq+', '')
+        (sink_url, sub_port) = url.split('+')
+        sink_port = sink_url.split(':')[-1]
+        sub_url = sink_url.replace(sink_port, sub_port)
+
+        sink = zmq.Context().socket(zmq.PUSH)
+        sink.connect(sink_url)
+
+        sub = zmq.Context().socket(zmq.SUB)
+        sub.setsockopt_string(zmq.SUBSCRIBE, u'')
+        sub.connect(sub_url)
+
+        self.sink = sink
+        self.sub = sub
+        self.channel = channel
+        super(ZmqManager, self).__init__(channel=channel,
+                                         write_only=write_only)
+
+    def _publish(self, data):
+        pickled_data = pickle.dumps(
+            {
+                'type': 'message',
+                'channel': self.channel,
+                'data': data
+            }
+        )
+        return self.sink.send(pickled_data)
+
+    def zmq_listen(self):
+        while True:
+            try:
+                response = self.sub.recv(flags=zmq.NOBLOCK)
+                if response is not None:
+                    yield response
+            except zmq.Again:
+                time.sleep(0.5)
+
+    def _listen(self):
+        for message in self.zmq_listen():
+            if isinstance(message, str):
+                message = pickle.loads(message)
+            if isinstance(message, dict) and \
+                    message['channel'] == self.channel and \
+                    'data' in message:
+                yield message['data']
+        return
--- a/socketio/zmq_manager.py
+++ b/socketio/zmq_manager.py
@@ -104,6 +104,7 @@ class ZmqManager(PubSubManager):  # pragma: no cover
             if isinstance(message, str):
                 message = pickle.loads(message)
             if isinstance(message, dict) and \
+                    message['type'] == 'message' and \
                     message['channel'] == self.channel and \
                     'data' in message:
                 yield message['data']
--- a/socketio/zmq_manager.py
+++ b/socketio/zmq_manager.py
@@ -1,9 +1,8 @@
-import time
 import pickle
 import re
 
 try:
-    import zmq
+    import eventlet.green.zmq as zmq
 except ImportError:
     zmq = None
 
@@ -92,12 +91,9 @@ class ZmqManager(PubSubManager):  # pragma: no cover
 
     def zmq_listen(self):
         while True:
-            try:
-                response = self.sub.recv(flags=zmq.NOBLOCK)
-                if response is not None:
-                    yield response
-            except zmq.Again:
-                time.sleep(0.5)
+            response = self.sub.recv()
+            if response is not None:
+                yield response
 
     def _listen(self):
         for message in self.zmq_listen():
--- a/socketio/__init__.py
+++ b/socketio/__init__.py
@@ -3,10 +3,11 @@ from .base_manager import BaseManager
 from .pubsub_manager import PubSubManager
 from .kombu_manager import KombuManager
 from .redis_manager import RedisManager
+from .zmq_manager import ZmqManager
 from .server import Server
 from .namespace import Namespace
 
 __version__ = '1.6.2'
 
 __all__ = [__version__, Middleware, Server, BaseManager, PubSubManager,
-           KombuManager, RedisManager, Namespace]
+           KombuManager, RedisManager, ZmqManager, Namespace]


etseidler:
@miguelgrinberg I'm not considering this a finished work, but I wanted to get your feedback to see if my approach would be considered general enough to eventually be merged in. there's another pull request for flask-socketio.

from looking at the Travis output, apparently there were some linting failures and perhaps some small incompatibilities with Python 3. i'll start looking at those.

miguelgrinberg:
Everything looks okay overall, but now going into the implementation details, maybe this is because I'm not familiar with zmq, but I don't really understand how the `jumpstart_subscription` method works. You are sending a packet with a type `subscribe`, but I don't see where is that packet received and handled.

etseidler:
as far as where the packet is received and handled, all messages that are
sent to the sink are handled by the external zmq message broker, which has
to be running in a separate process. i didn't include that message broker
in the pull request except here
<https://github.com/etseidler/python-socketio/commit/a37f6d13af528349a91e4816e0084ae08ae6c75f#diff-92b0b81caf2cf586f90ab5f5a5afa89bR35>
in the comments section of the zmq manager. the notion of having a separate
zmq message broker running is parallel to how a redis server is required to
use redis as a transport. the zmq message broker happens to be a little
more light-weight.


regarding the jumpstart_subscription method, a similar operation is
performed in redis_manager, but it is abstracted away by this line
<https://github.com/miguelgrinberg/python-socketio/blob/master/socketio/redis_manager.py#L65>.
That line leads to a subscribe command that is executed inside the redis
client itself, which can be seen here
<https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L2457>.

however... after taking another look at the jumpstart_subscription code,
i'm beginning to think it may be completely unnecessary for the zmq
transport. the subscribe message is received by the broker and then sent
back out to subscribers. the message broker doesn't do anything special
with that subscription message. i'll continue to look into this.

miguelgrinberg:
> the message broker doesn't do anything special with that subscription message.

Exactly, that's what I noticed. You are sending a message that nobody knows how to interpret.

etseidler:
i removed the logging and the unnecessary jumpstart_subscription method.i considered testing out the implementation of the zmq_manager in another app that wasn't my own.

etseidler:
i removed the `jumpstart_subscription` code as well as the logging code. as a next step, i was hoping to test out the use of the `ZmqManager` in an app that wasn't my own and i figured i would start with flack. i'll take a look at that next, but let me know if you have other ideas for testing/trying out zmq as a message transport.

socketio/zmq_manager.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+    while True:
+        publisher.send(receiver.recv())
+    ``
+
+    """
+    name = 'zmq'
+
+    def __init__(self, url='tcp://localhost:5555+5556',
```
miguelgrinberg: this default value should have the `zmq+` prefix.

socketio/zmq_manager.py
```diff
@@ -0,0 +1,110 @@
...
+            except zmq.Again:
+                time.sleep(0.5)
+
+    def _listen(self):
+        for message in self.zmq_listen():
+            if isinstance(message, str):
+                message = pickle.loads(message)
+            if isinstance(message, dict) and \
+                    message['channel'] == self.channel and \
+                    'data' in message:
```
miguelgrinberg: Don't you need to also check that the `type` is set to `message` here? If anything, as a precaution in case other message types are added in the future.

socketio/zmq_manager.py
```diff
@@ -0,0 +1,110 @@
...
+                response = self.sub.recv(flags=zmq.NOBLOCK)
+                if response is not None:
+                    yield response
+            except zmq.Again:
+                time.sleep(0.5)
+
+    def _listen(self):
+        for message in self.zmq_listen():
+            if isinstance(message, str):
+                message = pickle.loads(message)
```
miguelgrinberg: Maybe I am wrong, but it appears to me that when using this queue, each listener will receive every single message posted on the queue, not just the ones that come in the channel of interest.

If that is correct, then I think you should ensure that the unpickling here does not crash the script. If any unrelated message that is received is actually not pickled, then this will raise an exception.

miguelgrinberg:
@etseidler This is looking pretty good. I just wrote a few minor implementation notes. If you address those I think this is getting pretty close to done. Let me know how the testing goes.

etseidler:
thanks for the feedback! i'll keep you updated.

etseidler:
i was able to get ZmqManager working with the Flack app. alright! since the original flack app was set up with celery integration, i had to adjust it a little bit in order to get it working. i'd like to send you the exact changes i made to the flack app, so you can see them for yourself. should i submit a pull request to flack, just as a mechanism for showing the changes? i don't expect the changes to be merged in. if you have other ideas for a simple way to send you the minor flack changes that includes a diff view, please let me know.

by the way, i may not be able to send the flack changes until thursday or friday night.

miguelgrinberg:
@etseidler fork the flack repo, put your changes in a branch, then send me the repo and branch name. You can create a PR if you like, but it's not necessary.

etseidler:
@miguelgrinberg the changes i made to flack to get zmq working as a message queue transport are [here](https://github.com/etseidler/flack/commits/zmq-transport-mods).

after taking another look at those changes, i'm not totally convinced that the socketio emits are now being done on a separate process. however, that was not my main goal in testing using flack. it was to demonstrate that zmq could work as a message queue transport for apps besides the one that i am working on. please let me know what you think. i know i have emits working in distinct processes within my application.

one neat thing that i did was the following: got the flack app running, stopped the zmq message broker temporarily (mainly to show that the messages really were being sent via zmq), sent a few messages from multiple clients while the broker was down, and then started the message broker up again. with zmq's implementation, those messages sent while the broker was down were not lost, but were automatically picked up and sent out to the clients once the message broker was running again. that was pretty cool to see.

miguelgrinberg:
The changes that you made to Flack are minor. I left a couple of comments on your branch. Overall I feel pretty good, I think we will be able to merge this. :)

etseidler:
sweet! yeah, i was hoping that those changes to flack weren't that big of a deal.

etseidler:
just wanted to ensure my response from [the flack fork comments](https://github.com/etseidler/flack/commit/f40a463a2b5736f71b4beaf117f9228f68c9267d) gets captured in this thread...

> etseidler:
> feel free to make whatever final changes you'd like. that would be great.

miguelgrinberg:
@etseidler from your branch, basically the only thing I took is the ability to configure the Celery and Socket.IO queues separately. This is now committed to the Flack repository (you will need to checkout a brand new copy, or use git fetch and git reset to upgrade, as I inserted the change in the correct spot in the commit history). So now you can set your zmq URL on the `SOCKETIO_MESSAGE_QUEUE` environment variable.

I think that should be the only necessary change. I will test here in the next couple of days and if everything goes well I'll merge your PRs on this and the Flask-SocketIO repos. Thanks again!

etseidler:
hi @miguelgrinberg. have you had any time to do the testing you were wanting to do before the merge of this pull request?

miguelgrinberg:
@etseidler I'm finally having some time to look at this PR. Sorry for the delay.

Unfortunately I haven't been able to get Flack to work with these updates. I need to debug and try to figure out what's going on.

miguelgrinberg:
@etseidler I was finally able to get this PR to work for me. There was a small problem with `bytes` vs. `str`. I assume you were using Python 2 and because of that everything worked for you. I had to make a minor change for Python 3 to work as well.

After my testing, I came up with two potential problems.

1. The code in the `ZmqManager.zmq_listen()` constantly spins at a rate of 0.5s, instead of sleeping until there is something to read. The spinning is a problem in itself, since it causes some CPU usage while nothing is happening, but more concerning than that is the hardcoded 0.5 second sleep, which means that a message that arrives can be delayed for up to half a second until it is received. I think that `recv` call should block until there is a message, but unfortunately the blocking call appears to block eventlet. I wonder if this is the reason why eventlet offers its own green thread wrapper to zmq.

2. I find the little message broker that you proposed to not be as robust as I would like. Do you know of any zmq message broker project that can be recommended? Clearly most apps do not use a centralized message broker with zmq. I feel there must be a commonly accepted pattern for multi-publisher/multi-subscriber to work well together without having to rely on a custom written message repeater. Let me know if you can suggest anything, I haven't found anything yet, but I'll keep looking.

etseidler:
i completely agree with your assessment of the first potential problem. when i first started adapting the zmq manager, i struggled with this issue for quite a while. the no block flag was not ideal, but it at least provided a way forward. thank you so much for mentioning evenlet's green wrapper for zmq! i used that in place of vanilla zmq and it worked like a charm. i pushed a commit for that change.

yes, i'm using python > 2.7.9, so hopefully i can make any necessary changes to be compatible with python 3. was it the use of str in [this line](https://github.com/miguelgrinberg/python-socketio/pull/64/commits/88f3b87efa9f91d1e6eb23a110962d8d664ff1c3#diff-92b0b81caf2cf586f90ab5f5a5afa89bR104) in particular, or something else?

finally, regarding the message broker that i put together... there is [one project here](https://github.com/zeromq/malamute) that may be in line with what you're looking for. i don't know much about it, but it seems like it may be overkill, at least it would be for my app.
furthermore, while i am no expert on zmq, everything that i have read suggests that zmq eliminates the need for a central broker and if you do happen to need one, you can write your own without too much trouble. see [common zmq patterns](http://learning-0mq-with-pyzmq.readthedocs.io/en/latest/pyzmq/patterns/patterns.html), [broker vs brokerless](http://zeromq.org/whitepapers:brokerless), and [messaging patterns](http://zguide.zeromq.org/page:all#Messaging-Patterns), for example.
admittedly, i don't know the needs of other applications, but i would ask you what specifically is lacking in the simple message repeater/broker? is it performance concerns? custom behavior? architectural flaw? then, at some point, perhaps making a robust zmq message broker becomes a concern that is outside the scope of both flask-socketio and python-socketio. that is, any zmq message broker used would be a necessary part of the system, but doesn't seem to belong in either of the aforementioned repos/projects.

thank you once again for your feedback.

miguelgrinberg:
> everything that i have read suggests that zmq eliminates the need for a central broker

Right. Yet we are using one, and with my limited understanding of zeromq I don't see how you could do away with it.

> i would ask you what specifically is lacking in the simple message repeater/broker?

Robustness. For a hobby project it's totally fine, but I can't really recommend a solution that involves the developer copying a few lines of code and manually running a script. If the script were to crash the whole thing goes down. Redis and RabbitMQ install as services, the OS ensures these are always running.

> perhaps making a robust zmq message broker becomes a concern that is outside the scope of both flask-socketio and python-socketio

I think it is, but if there is no existing project that can do this, that is a problem.

etseidler:
> Right. Yet we are using one, and with my limited understanding of zeromq I don't see how you could do away with it.

To clarify, I believe that within this context (using zmq as a transport to facilitate separate process socketIO emits) the central broker is necessary. However, in the zmq community at large, there may not be a common demand for a central broker. Hence the lack of standard/accepted production-ready central brokers. Again, malamute may be the exception if it ends up being what we want.

> For a hobby project it's totally fine, but I can't really recommend a solution that involves the developer copying a few lines of code and manually running a script.

Thank you for helping me to understand your concern about robustness. I can see how this setup is less than ideal for a production environment. Other than malamute or possibly wrapping the Python script in some type of service, I don't currently have a good way to address this concern. You already somewhat alluded to this, but if you were willing to commit the zmq transport changes as an experimental feature, perhaps someone else would come along later and suggest a better solution for the central broker.


> in my view, it is really not that useful if there is no clear way to connect many publishers with many subscribers

I don't understand this concern yet. Let me share what I tend to think of in response to this; please correct me if I'm off. I think of the fact that each instance of the zmq manager uses a push socket to push all messages to a central broker. the broker uses a pull socket on port A to gather pushes from all instances of zmq manager, then publishes those messages out to anyone who is listening on port B. all zmq manager instances happen to be subscribed to messages on port B, so they receive all messages. so while the zmq manager instances aren't technically using a pub socket, each instance is sending out messages and also receiving all messages that are sent out. to me, this seems to have the effect of being a multi-pub/multi-sub.

etseidler:
it just occurred to be that maybe it is possible to eliminate the central broker completely. is that what you were referring to with the multi-pub/multi-sub comments?

with the current implementation of ZmqManager, all messages from ZmqManager instances go to the central broker and from there back out to the instances. it may make more sense to have each instance of the ZmqManager send messages out as individual publishers and also receive messages as individual subscribers. If that works (I'm not certain at this point if it will, I'll have to try it), then it would seem that the central broker would not be needed at all. if i'm understanding it correctly, this would completely side-step the need for a robust broker because no broker would be needed. and this would also more closely align with the 'no-broker' philosophy of zmq. in fact, the only reason i used a broker to begin with was because i was modeling the zmq manager after the redis manager. there may be some issues that make this impractical, but i will investigate.

miguelgrinberg:
Based on what I read, you still need some sort of central entity that keeps track of all the nodes. If not, how can I subscribe know where are all the publishers? And this also needs to be dynamic, as nodes can come online or go offline at any time.

socketio/zmq_manager.py
```diff
@@ -0,0 +1,110 @@
...
+                which will need to be provided and running.
+    :param channel: The channel name on which the server sends and receives
+                    notifications. Must be the same in all the servers.
+    :param write_only: If set to ``True``, only initialize to emit events. The
+                       default of ``False`` initializes the class for emitting
+                       and receiving.
+
+
+    N.B.
+    a zmq message broker must be running for the zmq_manager to work.
```
kc64: There are no brokers in the classical sense in zmq.

socketio/zmq_manager.py
```diff
@@ -0,0 +1,110 @@
...
+        sink_port = sink_url.split(':')[-1]
+        sub_url = sink_url.replace(sink_port, sub_port)
+
+        sink = zmq.Context().socket(zmq.PUSH)
+        sink.connect(sink_url)
+
+        sub = zmq.Context().socket(zmq.SUB)
+        sub.setsockopt_string(zmq.SUBSCRIBE, u'')
+        sub.connect(sub_url)
+
```
kc64: I'd recommend the following to make the code more robust and readable.  The atomic regular expression operations prevent the need for operating on various chunks.  Also, the more important part is the use of PUB rather than PUSH.  PUSH will distribute requests unequally among PULLers while a PUB on bind will distribute equally among all SUBs.

`r = re.compile('^zmq\+(?P<protocol>\w+)://(?P<tld>\w+):(?P<port1>\d+)\+(?P<port2>\d+)$')
tokens = r.match(url).groupdict()

sink = zmq.Context().socket(zmq.PUB)
sink.connect('{protocol}://{tld}:{port1}'.format(**tokens))

sub = zmq.Context().socket(zmq.SUB)
sub.setsockopt_string(zmq.SUBSCRIBE, u'')
sub.connect('{protocol}://{tld}:{port2}'.format(**tokens))
`

socketio/zmq_manager.py
```diff
@@ -0,0 +1,110 @@
...
+
+    N.B.
+    a zmq message broker must be running for the zmq_manager to work.
+    you can write your own or adapt one from the following simple broker below.
+    port numbers in the broker must match port numbers in connection string.
+    ``
+    import zmq
+
+    receiver = zmq.Context().socket(zmq.PULL)
```
kc64: `receiver = zmq.Context().socket(zmq.SUB)
receiver.setsockopt(zmq.SUBSCRIBE, '')
`

miguelgrinberg:
@kc64 Thanks for your review. Unfortunately I'm not experienced enough with ZMQ to comment. This code was contributed, and as it is committed I have verified works fine. If you can propose a way to eliminate the broker, I would be interested in that.

> the more important part is the use of PUB rather than PUSH. PUSH will distribute requests unequally among PULLers while a PUB on bind will distribute equally among all SUBs.

As I said above, I did not write this code, but I'm not sure your comment applies, since the PUSH/PULL connection is always a one-to-one. The broker script is a publisher, and all the server nodes are subscribers, and also have a pull/push connection to the broker, though which they send to the broker what needs to be published. The structure makes sense to me, but if you can propose a better way, I would definitely consider it.

kc64:
> If you can propose a way to eliminate the broker, I would be interested in that.

I think you're stuck with it.  If many instances of socketio/ZMQManager can be made (I'm asserting they can), and since the zmq PUB/SUB distribution pattern only works in 1-to-many fashion if the single-point entity _binds_ the socket and the many points _connect_ to the socket, then the ZMQManager instances must always do the _connecting_ on both PUB and SUB.  The "broker" is not a broker but a stub that fills in for the broker required in other platforms like Redis.

> the PUSH/PULL connection is always a one-to-one

Not according to [the zmq documentation](http://zguide.zeromq.org/py:all#Divide-and-Conquer). The figure and text demonstrate that a PUSHer distributes messages (like a card dealer with cards) to whichever PULLers are available to accept them.  If you are attempting to communicate in a manner wherein all clients receive all messages, then this pattern will not work.

BTW, I've tested the PUB/SUB pattern in place of this PUSH/PULL implementation and it works fine.

miguelgrinberg:
> the PUSH/PULL connection is always a one-to-one
> > Not according to the zmq documentation.

I don't mean in genenral. I mean in how @etseidler set this up for this particular case. The communication over PUSH/PULL always has one sender and one receiver.

> BTW, I've tested the PUB/SUB pattern in place of this PUSH/PULL implementation and it works fine.

What's the advantage, when there is one sender and one receiver?

etseidler:
in this case, it seems that we're stuck with a central broker. all PUSHes on the ZmqManager instances will be received by a central PULL socket that sits on the broker. while the socket communication from the ZmqManager instances to the broker _could_ be done via PUB/SUB, I agree that there doesn't seem to be any benefit to having it set up that way. in fact, PUSH/PULL seems _more_ appropriate since all the PUSHed messages will be going to one place, the central broker. 
pushpull

PR #66, opened 2017-01-22
Fix namespace hyphens

Kurlov:
Hello
There is [an issue](https://github.com/miguelgrinberg/Flask-SocketIO/issues/392) with hyphens namespace in Flask-SocketIO repo which is actually python-socketio bug. I think this PR will fix it. I also briefly compared decode time with these changes and looks like it performs a little bit faster.
Please, review. 
Thank you

### DIFF
--- a/socketio/packet.py
+++ b/socketio/packet.py
@@ -78,12 +78,21 @@ class Packet(object):
         self.namespace = None
         self.data = None
         ep = ep[1:]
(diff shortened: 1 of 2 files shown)


miguelgrinberg:
Thanks!

I reviewed your fix and found it to have a problem with packets that have to data payload, which is something that is not common, but possible. So I adapted the fix to work in all cases before merging it.

PR #91, opened 2017-03-31
aiohttp instead of flask in example HTML

jacopofar:
I tried the example chat application and was confused by the _Flask_ reference. It's a refuse, this fixes it :)


miguelgrinberg:
The title of the page should say `python-socketio`, as that is the name of this package. I missed updating it when I ported this from the Flask specific package.

Can I ask you to use this name, and also update the other examples in the same way? Thanks!

jacopofar:
OK!

PR #103, opened 2017-05-23
The usage of class-based namespace

Stonesjtu:
There should be an additional `self` argument passed to 
each member function of custom Namespaces.


miguelgrinberg:
Thank you!

PR #116, opened 2017-07-14
redis pub/sub support set password

larry-dev:
At first I added it to URL, but it was not easy to handle when the password value was None, so I added a parameter "password"

### DIFF
--- a/socketio/asyncio_redis_manager.py
+++ b/socketio/asyncio_redis_manager.py
@@ -53,7 +53,7 @@ class AsyncRedisManager(AsyncPubSubManager):  # pragma: no cover
     name = 'aioredis'
 
     def __init__(self, url='redis://localhost:6379/0', channel='socketio',
-                 write_only=False):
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
I'm in the middle of a long trip back home today, I'll review this in a day or two. Thanks!

larry-dev:
sorry,When do you plan to join this function?

miguelgrinberg:
@larry-dev did you notice that the automated build is failing with your changes? Do you have time to fix those problems, or do you want me to? I can't merge this as is.

larry-dev:
i fixed those problems.sorry,i didn't notice this error of build

miguelgrinberg:
@larry-dev I modified your change to include the password in the URL, so that it is consistent with the non-async side. To add a password, use `redis://:password@hostname:6379/0`. Note that colon preceding the password, this is required per URL syntax.

PR #154, opened 2017-12-18
added kwargs to the __init__ of RedisManager

IddoPuff:
added kwargs to the __init__ of RedisManager in order to pass more parameters
in this case used to optionally set redis client name
add kwargs to zmq_manager.py and kombu_manager.py

### DIFF
--- a/socketio/kombu_manager.py
+++ b/socketio/kombu_manager.py
@@ -38,7 +38,7 @@ class KombuManager(PubSubManager):  # pragma: no cover
     name = 'kombu'
 
     def __init__(self, url='amqp://guest:guest@localhost:5672//',
-                 channel='socketio', write_only=False):
+                 channel='socketio', write_only=False, **kwargs):
         if kombu is None:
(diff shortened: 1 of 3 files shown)


miguelgrinberg:
The Redis constructor takes `kwargs` too, so I think if `RedisManager` is going to take `kwargs` those should be passed-through to Redis.

Repurposing the `kwargs` to start adding arbitrary commands such as the client_setname does not seem like a good idea. I think I would prefer that `RedisManager` optionally accepts an already configured `Redis` instance, so that you can provide any customizations that you need.

IddoPuff:
I see your point and changed the code to reflect it. let me know what you think if it

miguelgrinberg:
Providing a connection instance is not a good way to do this, because it would prevent the server from reconnecting if the original connection is dropped, for example as a result of a redis server restart.

It would be better to provide a connection function that can be called to obtain a connection. I'm going to close this PR, since there is no way to make it work alongside the redis reconnection features currently in place.

PR #184, opened 2018-05-21
Compatibility: README update

kaisbe:
It's not compatible with Python 2 and 3.3, because of the used `async` / `await` syntax that's only available in Python 3.5+.


miguelgrinberg:
You are incorrect, this package is compatible with 2.7 and 3.3. Using asyncio is optional, not required.

damoit:
Latest python-socketio requires Python 3.6+ .

Chapter _Client features_ in README says: _Compatible with Python 3.5+_

miguelgrinberg:
@damoit Fixed. Thanks for the report.

PR #197, opened 2018-07-25
Logging improvements for write-only connections

truetech:
While trying to emit from the standalone, write-only `KombuManager` (using the example from the docs), I kept running into this exception:

```
AttributeError: ‘NoneType’ object has no attribute ‘logger’
```

It looks like logging inside all of the client managers depends on the logger created when the server was instantiated, but that is `None` when the manager is used by itself in write-only mode. The test for write-only sets a mock server (and thus a default logger), so it doesn't fail in the tests.

This code should address that, creating a new option to set the logger for write-only managers, using the server logger if it exists, and falling back to the same default logger as the server. It also fixes what looks like a bug where `KombuManager` was not passing `write_only` up to the `PubSubManager` and adds a test to ensure loggers are being set. Logging in `AsyncRedisManager` was done a little differently; I'm not sure if that was intentional (no option, always default logger), but I made it consistent with the others.

### DIFF
--- a/socketio/asyncio_manager.py
+++ b/socketio/asyncio_manager.py
@@ -44,7 +44,7 @@ class AsyncManager(BaseManager):
             callback = self.callbacks[sid][namespace][id]
         except KeyError:
             # if we get an unknown callback we just ignore it
-            self.server.logger.warning('Unknown callback received, ignoring.')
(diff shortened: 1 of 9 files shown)


miguelgrinberg:
Thanks, this is a nice improvement!

PR #200, opened 2018-08-15
wip: Sanic + NATS example with async pub sub manager

wallyqs:
Hello :) Some NATS users requested for an example of using asyncio socket.io and the NATS asyncio client together and want to help so gave it a shot at adapting one of the examples for use with NATS (https://github.com/nats-io/asyncio-nats/issues/53)

I am not very familiar yet with how to properly use the framework so looking for some feedback... Have made the following work with Sanic async mode so sharing, it subscribes to the `socketio` subject by default to receive events and can also publish NATS events via the browser.  One thing that I could not figure out is how to make the browser receive the events that have been received via the AsyncPubSubManager.

To try the example, first needs to install the following deps:

```
pip install asyncio-nats-client

# Latest NATS server (with debug + tracing for dev)
# Releases available at: https://github.com/nats-io/gnatsd/releases
gnatsd -DV

cd examples/sanic
python3 nats-app.py 
```

Thanks!

### DIFF
--- a/examples/sanic/nats-app.py
+++ b/examples/sanic/nats-app.py
@@ -1,0 +1,103 @@
+import asyncio
+
+from sanic import Sanic
+from sanic.response import html
+
+import socketio
+
+# --- NATS PubSub Manager ---
+
+import nats
+from nats.aio.client import Client as NATS
+from socketio.asyncio_pubsub_manager import AsyncPubSubManager
+
+import json
+
+class AsyncNatsManager(AsyncPubSubManager):
+    name = 'asyncionats'
+
+    def __init__(self,
+                 servers=None,
+                 channel='socketio',
+                 write_only=False,
+                 loop=asyncio.get_event_loop(),
+                 ):
+
+        if servers is None:
+            servers = ["nats://127.0.0.1:4222"]
+        self.servers = servers
+        self.queue = asyncio.Queue()
+
+        # Establish single connection to NATS for the client.
+        self.nc = None
+        super().__init__(channel=channel, write_only=write_only)
+
+    async def _publish(self, data):
+        print("Socket.io <<- NATS     :", data)
+
+        # Send the client events through NATS
+        if self.nc is None:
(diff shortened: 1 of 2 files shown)


miguelgrinberg:
Not sure I understand the goal that you have. The `AsyncPubSubManager` class is used internally when multiple servers need to coordinate their actions. More importantly, this class has not interaction with clients (browsers or otherwise), it is only used for server to server communication.

If you explain what you want to achieve by combining Socket.IO and NATS maybe I can advice you on an implementation. Unless I'm missing something, I believe the approach that you've taken is flawed.

wallyqs:
@miguelgrinberg thanks for the feedback, this is helpful. I'm trying to understand what are the responsibilities of the `AsyncPubSubManager` or how users might be trying to adopt this library... NATS is just also just pure pub/sub so in terms of usage it would be analogous to using the `RedisManager` implementation as well but think I misunderstood what it should do.

Is there a way to receive events from a pub/sub backend and then have them be emitted via socket.io events to the browser?  Thanks again for taking a look.

wallyqs:
(closing for now)

miguelgrinberg:
@wallyqs This library implements the server part of the Socket.IO protocol. It's not a pub/sub library, but more of a general event dispatching library that uses HTTP and WebSocket as transport mechanisms. Pub/sub functionality could be implemented on top of Socket.IO.

In the most simple deployment case, you have a Socket.IO server and that's it. Socket.IO clients running on browsers, smartphones, etc. can connect and start sending and receiving asynchronous events. No need for redis, or anything else.

If you need to scale out your server to handle larger number of concurrent clients, then you can run multiple server processes. These server processes communicate among themselves through a pub/sub queue. This is exclusively internal communication between servers to coordinate actions such as broadcasting of events.

So really I don't see what could be a useful integration between NATS and Socket.IO, since the pub/sub mechanisms in this server are not accessible to clients.

PR #221, opened 2018-12-10
Add shutdown function from engine io async server.

Viatorus:
Depends on:
https://github.com/miguelgrinberg/python-engineio/pull/85


miguelgrinberg:
This issue will be automatically closed due to being inactive for more than six months. Please reopen if you need more assistance.

Issue #227, opened 2018-12-29
How to close a client?


davidkingzyb:
Hello,
I use flask-socketio and python-socketio[client].
After I disconnect from the client the script is still runing and ping pong with the server.
How to close the client?

miguelgrinberg:
How do you disconnect? If you are calling the `wait()` method in the client, the expected behavior is that when the connection breaks this function should return, but this is all very new code, maybe I missed some use case.

davidkingzyb:
I tried calling the `wait()` method in the client, but after called `disconnect()` the connection was not broken. Is there a `destory()` method here? I can use it to cleanup the client.
```
    sio=socketio.Client()
    @sio.on('connect')
    def on_connect():
        print('disconnet')
        sio.disconnect()
    sio.connect(conf.IO_HOST)   
    sio.wait()
```

miguelgrinberg:
You can't call disconnect in the connect handler, the connection isn't established yet at that point. Return `False` from the handler if you want to reject the connection. This is covered in the docs, by the way.

davidkingzyb:
I solved my problem by calling `sio.eio.disconnect(True)` after `sio.disconnect()`. I think maybe this step should be done in `disconnect()`? I'm not sure it's the right way, please tell me. Thanks for your help and Happy New Year !  🎉

davidkingzyb:
#228

saltaverde:
@davidkingzyb FYI I was having a problem in my application's unit tests where after I called `sio.disconnect()` and the test ended I was seeing `BrokenPipeError` exception on a thread other than the main one. I added `sio.eio.disconnect()` and that seems to remedy the problem.

davidkingzyb:
Maybe the old version has this problem. This bug should have been fixed at v3.0.1 #228, now `sio.disconnect()` contain `eio.disconnect()`, you don't need call it manually.

saltaverde:
Hmm, I'm using v3.1.2 and was experiencing the `BrokenPipeError` when not manually calling `sio.eio.disconnect()`

neuronflow:
Hmm I  am facing the exact same problem....things I tried so far:

`              self.sio.disconnect()
                print("**************")
                print("disco called")
                self.sio.eio.disconnect()
                print("sio.eio called")
                sys.exit(0)
                print("exit called")
                return`

using `python-engineio==3.10.0
python-socketio==4.4.0`

miguelgrinberg:
[...]

PR #228, opened 2018-12-31
disconnect engineio when call disconnect

davidkingzyb:
solve the problem miguelgrinberg/python-socketio#227 after call `disconnect` disconnect engineio too

### DIFF
--- a/socketio/client.py
+++ b/socketio/client.py
@@ -293,6 +293,7 @@ class Client(object):
         self._trigger_event('disconnect', namespace='/')
         self._send_packet(packet.Packet(
             packet.DISCONNECT, namespace='/'))
+        self.eio.disconnect(True)
 
     def transport(self):
         """Return the name of the transport used by the client.


miguelgrinberg:
Hmm. Not sure this is right. I'll have to check with the JavaScript server, but my understanding is that a disconnect is issued with the `DISCONNECT` packet. At the Socket.IO level there are multiple logical connections all multiplexed over the same Engine.IO connection, what you are doing is going to break all connections, not just the one you call disconnect on.

davidkingzyb:
Maybe I am thinking wrong. When I need another connection should I new a `socketio.Client()`? the eio does not look like a singleton. how to share the connection?

I have checked javascript client, `disconnect()` and `close()` is  same method, the method send `packet.DISCONNECT` and destroy the manager and close the engineio. I'm not sure this client should has same behavior and usage in same way. 

Could you give an example with disconnect and multiple connections? Thank you very much.

miguelgrinberg:
Thanks. I have looked at this problem and in the end I think your solution is appropriate. I have committed it along with a similar fix for the asyncio client, plus unit tests.

PR #247, opened 2019-02-04
pass 'binary' flag into AsyncServer constructor

maksutovaynur:
There was no chance to support binary payloads because 'binary' flag cannot be passed twice into constructor


miguelgrinberg:
As I said in the issue, the `binary` flag doesn't do anything in Python 3, it is completely ignore. Since the async server is Python 3 only, this setting is misleading to have.

PR #269, opened 2019-03-12
Add namespaces parameter to the self.connect call during reconnection

Keylekan:


miguelgrinberg:
Good catch! I merged this PR manually with a minor change, and also added the same change to the asyncio client. Thanks!

PR #289, opened 2019-04-19
Add missing import statement for socketio in docs

faheel:


miguelgrinberg:
@faheel Thanks. The error is actually that I used `import engineio` instead of `import socketio`. Can you please remove the engineio import?

faheel:
@miguelgrinberg Done.

miguelgrinberg:
Thank you!

PR #319, opened 2019-07-19
Added rediss:// URL scheme to AsyncRedisManager

### LINKED ISSUE #318: AsyncRedisManager SSL Support
Is there a way to use SSL with the AsyncRedisManager?  This is possible with RedisManager by passing in a "rediss" scheme instead of a "redis" scheme, but I don't see a way to do it with the Async manager.

I have sort of patched in a way to do this just by adding an "ssl" option to `AsyncRedisManager.__init__()`, then passing it through to `aioredis.create_redis()` in `_publish` and `_listen`.  Happy to do a pull request if that's the right way to do this.

dbanty:
Closes #318

### DIFF
--- a/socketio/asyncio_redis_manager.py
+++ b/socketio/asyncio_redis_manager.py
@@ -56,7 +56,9 @@ class AsyncRedisManager(AsyncPubSubManager):  # pragma: no cover
             raise RuntimeError('Redis package is not installed '
                                '(Run "pip install aioredis" in your '
                                'virtualenv).')
-        self.host, self.port, self.password, self.db, self.ssl = _parse_redis_url(url)
+        (
+            self.host, self.port, self.password, self.db, self.ssl
+        ) = _parse_redis_url(url)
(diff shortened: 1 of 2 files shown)


miguelgrinberg:
Two comments:
- Could you fix the error from the build?
- Any chance you can also do the standard (i.e. non-asyncio) Redis manager?

Thanks!

dbanty:
> * Could you fix the error from the build?

Done I think, waiting on Travis.

> * Any chance you can also do the standard (i.e. non-asyncio) Redis manager?

The standard manager already supports rediss I think.  It uses redis.Redis.from_url which supports the scheme ( see [here](https://github.com/andymccurdy/redis-py/blob/6e23760fefbdf27ec941a2c3ae1ec2657875ac31/redis/client.py#L605) )

miguelgrinberg:
Ay yes, the URLs are fully handled by the Redis client for the non-asyncio case. Great.

Issue #347, opened 2019-09-04
[Question] django_example url question.


625781186:
[socketio_app/management/commands/runserver.py](https://github.com/miguelgrinberg/python-socketio/blob/HEAD/examples/server/wsgi/django_example/socketio_app/management/commands/runserver.py) 
error 1. It lead to `django admin` lost css style ( commented the url); 

![jdfw11](https://user-images.githubusercontent.com/30521429/64293265-cea1bf80-cf9e-11e9-8f4c-6da7abdb9ef4.gif)

error 2. It lead to can't open other url (not commented the url) , just jump to `index.html` only.

![jdfw22](https://user-images.githubusercontent.com/30521429/64293284-d5c8cd80-cf9e-11e9-8caf-b38de10838ab.gif)

Make sure it's the `runserver.py` problem(use python manage.py runserver **--noreload** after rename `runserver.py` In order to avoid `ValueError: signal only works in main thread`):

![jdfw33](https://user-images.githubusercontent.com/30521429/64293299-dc574500-cf9e-11e9-8eeb-1131a3b16436.gif)

I don't konw how to fix it , If you have time, please help to fix it.

625781186:
I solve error 2 like this :
`url(r'^$', views.do, name='index'),`
![image](https://user-images.githubusercontent.com/30521429/64294004-69e76480-cfa0-11e9-9d8b-47829a87b856.png)

but  error 1 still exit.

625781186:
I solve error 1 like this :
    https://stackoverflow.com/questions/34164799/custom-runserver-command-gets-404-on-static-files

PR #348, opened 2019-09-05
fix : static not found

625781186:
fix : static not found

### DIFF
--- a/examples/server/wsgi/django_example/socketio_app/management/commands/runserver.py
+++ b/examples/server/wsgi/django_example/socketio_app/management/commands/runserver.py
@@ -1,9 +1,10 @@
 from django.core.management.commands.runserver import Command as RunCommand
+from django.contrib.staticfiles.management.commands import runserver 
 
 from socketio_app.views import sio
 
 
-class Command(RunCommand):
+class Command(runserver.Command):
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
What version of django are you using. I see the import in the latest release: https://github.com/django/django/blob/master/django/core/management/commands/runserver.py#L23.

625781186:
> What version of django are you using. I see the import in the latest release: https://github.com/django/django/blob/master/django/core/management/commands/runserver.py#L23.

I use django==2.1.

this detail infos : https://github.com/miguelgrinberg/python-socketio/issues/347

miguelgrinberg:
Does this work when using eventlet or gevent async modes?

625781186:
> Does this work when using eventlet or gevent async modes?

no , I haven't made any other changes to the example( [django_example/socketio_app/views.py](https://github.com/miguelgrinberg/python-socketio/blob/a839a36fa0fa7f0e5d8976ff47b217f6b1e8a44b/examples/server/wsgi/django_example/socketio_app/views.py) ).

miguelgrinberg:
Tested latest version of Django and this problem does not exist anymore. Closing.

PR #427, opened 2020-02-12
fix the bug that socketio can't emit event disconnnect when call  disconnect() 

higrw123:
fix the bug that socketio can't emit event disconnnect when call  disconnect()

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -45,7 +45,7 @@ The example that follows shows a simple Python client:
 Client Features
 ---------------
 
-- Can connect to other Socket.IO complaint servers besides the one in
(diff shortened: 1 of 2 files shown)


miguelgrinberg:
Did you look at the history in git regarding that line you are commenting out? It was put there specifically to address an issue, you can't just comment it out. Also if you can explain more clearly what case you are trying to address I would appreciate it.

higrw123:
> appreciate
ok, i am do this first time, just try fix the problem when i use it, i will learn more information about how to address an issue

miguelgrinberg:
Please reopen once you can provide information on the issue you are trying to address. Thanks.

PR #488, opened 2020-05-20
AsyncPubSubManager does not await for can_disconnect

GitHK:
AsyncPubSubManager had an issue which was not caught by the `test_can_disconnect`.

- Fixed the test
- Fixed the issue


miguelgrinberg:
Thanks!

PR #493, opened 2020-05-26
implement uWSGI queue manager

johaven:
Do not hesitate to tell me if I forgot things. I can also do a PR on the flask-socketio side if necessary.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -108,6 +108,12 @@ API Reference
 .. autoclass:: RedisManager
    :members:
 
+``UWSGIManager`` class
+----------------------
+
+.. autoclass:: UWSGIManager
+   :members:
+
 ``KafkaManager`` class
 ----------------------
 
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -461,6 +461,31 @@ to connect to a message queue such as `Redis <http://redis.io/>`_ or
 `RabbitMQ <https://www.rabbitmq.com/>`_, to communicate with other related
 Socket.IO servers or auxiliary workers.
 
+uWSGI
+~~~~~
+
+To use a uWSGI message queue, you need to run uWSGI server with option --queue::
+
+    # One queue with 1 slot
+    uwsgi --http :5000 --gevent 1000 --queue 1 --http-websockets --master --wsgi-file app.py --callable app
+
+Only one slot is required because only the last message is dequeued.
+The uWSGI queue is processed when a uWSGI signal is triggered on a specific channel (default 0).
+
+The uWSGI queue is configured through the :class:`socketio.UWSGIManager`::
+
+    # socketio.Server class
+    mgr = socketio.UWSGIManager()
+    sio = socketio.Server(client_manager=mgr)
+
+If you already use uWSGI signals you can specify on which channel the events will be triggered::
+
+    # channel must be a non-negative integer
+    socketio.UWSGIManager('uwsgi:8')
+
+Note that uWSGI currently was not tested with asyncio, so it cannot be used with
+the :class:`socketio.AsyncServer` class.
+
 Redis
 ~~~~~
 
--- a/socketio/__init__.py
+++ b/socketio/__init__.py
@@ -7,6 +7,7 @@ from .kombu_manager import KombuManager
 from .redis_manager import RedisManager
 from .kafka_manager import KafkaManager
 from .zmq_manager import ZmqManager
+from .uwsgi_manager import UWSGIManager
 from .server import Server
 from .namespace import Namespace, ClientNamespace
 from .middleware import WSGIApp, Middleware
@@ -31,7 +32,8 @@ __version__ = '4.6.1dev'
 
 __all__ = ['__version__', 'Client', 'Server', 'BaseManager', 'PubSubManager',
            'KombuManager', 'RedisManager', 'ZmqManager', 'KafkaManager',
-           'Namespace', 'ClientNamespace', 'WSGIApp', 'Middleware']
+           'UWSGIManager', 'Namespace', 'ClientNamespace', 'WSGIApp',
+           'Middleware']
 if AsyncServer is not None:  # pragma: no cover
     __all__ += ['AsyncClient', 'AsyncServer', 'AsyncNamespace',
                 'AsyncClientNamespace', 'AsyncManager', 'AsyncRedisManager',
--- a/socketio/uwsgi_manager.py
+++ b/socketio/uwsgi_manager.py
@@ -31,7 +31,8 @@ class UWSGIManager(PubSubManager):  # pragma: no cover
     """
     name = 'uwsgi'
 
-    def __init__(self, url='uwsgi:0', channel='socketio', write_only=False, logger=None):
+    def __init__(self, url='uwsgi:0', channel='socketio', write_only=False,
+                 logger=None):
         self._check_configuration()
         self.signum = self._sig_number(url)
         self.queue = Queue()  # uWSGI does not provide a a blocking queue
@@ -46,7 +47,8 @@ class UWSGIManager(PubSubManager):  # pragma: no cover
         try:
             uwsgi.queue_last()
         except AttributeError:
-            raise RuntimeError('uWSGI queue must be enabled with option --queue 1')
+            raise RuntimeError('uWSGI queue must be enabled with '
+                               'option --queue 1')
 
     @staticmethod
     def _sig_number(url):
@@ -54,7 +56,8 @@ class UWSGIManager(PubSubManager):  # pragma: no cover
             try:
                 sig = int(url.split(':')[1])
             except ValueError:
-                logger.warning('Bad URL format %s, uWSGI signal is listening on default (1)' % url)
+                logger.warning('Bad URL format %s, uWSGI signal '
+                               'is listening on default (1)' % url)
             else:
                 return sig
         return 0
--- a/socketio/uwsgi_manager.py
+++ b/socketio/uwsgi_manager.py
@@ -64,7 +64,7 @@ class UWSGIManager(PubSubManager):  # pragma: no cover
         return 0
 
     def _publish(self, data):
-        logger.warning('pubish from worker %s' % uwsgi.worker_id())
+        logger.warning('publish from worker %s' % uwsgi.worker_id())
         uwsgi.queue_push(pickle.dumps(data))
         uwsgi.signal(self.signum)
 
--- a/socketio/uwsgi_manager.py
+++ b/socketio/uwsgi_manager.py
@@ -1,6 +1,5 @@
 import logging
 import pickle
-from queue import Queue
 
 try:
     import uwsgi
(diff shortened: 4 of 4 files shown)


johaven:
Flake8 failed, but it's not related to this PR.

johaven:
I adjusted a few things for me it's good :)

miguelgrinberg:
Sorry, but can you point me to documentation that shows how to use this uwsgi based queue for pub/sub? I believe you are using a standard queue, which is not going to work for this.

Also unclear on how this would work if you run multiple uwsgi processes.

johaven:
uWSGI does not support native pub/sub, I used 2 interesting frameworks to reproduce this.

uWSGI queue: https://uwsgi-docs.readthedocs.io/en/latest/Queue.html
uWSGI signals: https://uwsgi-docs.readthedocs.io/en/latest/Signals.html

Standard Python queue (as i said in doc string) is only used to make the _listen_ part blocking but messages are received from the uwsgi queue.

miguelgrinberg:
How does this work when I have two or more uwsgi workers running on different servers?

johaven:
From what I know it is not possible with uWSGI to run workers on different servers.

In the uWSGi language a worker is a process, often managed by the master process.

I mentionned in doc string: _This class implements a UWSGI backend for event sharing across multiple processes._

miguelgrinberg:
Yeah, unfortunately this does not work. I'm not sure how you made it work for you, because there is really no configuration in which this can be used.

The only two situations in which a pub/sub client manager makes sense is when running multiple servers, or when running auxiliary processes. This implementation supports neither.

johaven:
I'm not entirely agree with that.

Explain to me how the messages are sent between a client 1 on process 1 and a client 2 on a process 2 in a case of multiprocessing?

Each process / worker is independent at the memory stack level.

miguelgrinberg:
I don't understand the question. The function of the message queue is to pass messages between processes. All the implementations do this, and this is actually the only reason this project uses a message queue.

What part do you not agree with?

johaven:
My problem is to be able to manage my application on several workers / processes via UWSGI and to handle socketio traffic on all workers.

What I've observed so far is that the emit and received messages are process confined. It is not possible for user 1 on process 1 to send a message to user 2 on process 2.

This is why I propose this solution, but your remarks surprised me so I dug a little and it would seem that the listen method of pubsub_manager is called only in a single process (even if 4 processes are launched ); this is why it does not work properly (I was lucky during my tests to come across the same worker).

I logged the behaviour:
```
start listen on process 4
emitting event "check" to 2d97ac4e115c4236b6209bb6678b0d0e [/notifications]
publish from process 1
enqueue on process 2: {'method': 'emit', 'event': 'check', 'data': None, 'namespace': '/notifications', ...}
10.211.55.10 [pid: 28629|app: 0|req: 9/29] 10.211.55.2 () {52 vars in 2088 bytes} [Wed May 27 17:17:26 2020] POST /api/v1/comments/share/toot => generated 10 bytes in 30 msecs (HTTP/1.1 200) 2 headers in 71 bytes (3 switches on core 499)
enqueue on process 1: {'method': 'emit', 'event': 'check', 'data': None, 'namespace': '/notifications', ...}
enqueue on process 4: {'method': 'emit', 'event': 'check', 'data': None, 'namespace': '/notifications', 'room': '2d97ac4e115c4236b6209bb6678b0d0e', 'skip_sid': None, 'callback': None, 'host_id': 'a9e0590a7b084374a0aed5d1b8b9d607'}
iter from process 4: {'method': 'emit', 'event': 'check', 'data': None, 'namespace': '/notifications', ...}
send message from process 4
pubsub message: emit
enqueue on process 3: {'method': 'emit', 'event': 'check', 'data': None, 'namespace': '/notifications', ..}
2d97ac4e115c4236b6209bb6678b0d0e: Sending packet MESSAGE data 2/notifications,["check"]
```
Indeed in this case the queues of processes other than number 4 are never dequeued and the user who may be on the other processes will never receive messages.

The fact that listen to the pubsub manager is only triggered on a single process probably comes from the bootstrap of the flask application by uWSGI, do you confirm?

miguelgrinberg:
I continue to be confused about what you are saying. Have you read the docs regarding running multiple workers?

This is something that has been working for many years now, and as I said before, it is exactly the issue that the message queue addresses.

> It is not possible for user 1 on process 1 to send a message to user 2 on process 2.

This is supported when you connect all workers to the message queue.

johaven:
Yes i know, i already read all docs from flask-socketio, python-socketio :)

But why would I use an external message broker like Redis if I can (maybe) use uWSGI which already runs all of my code in production?

In terms of latency, I have more advantages in using uWSGI which allows me to communicate between my processes directly rather than going through a socket of an external broker.
In terms of supervision, it is one less component to manage.
I also use uWSGI for the cache part of my application, so if I can centralize everything it's not bad.

The question is rather why not? There is a redis broker, a zmq broker ... Why not use uWSGI directly since this is mainly what we all use to host our applications ?

I find that my idea is not so bad, in terms of integration we would have support for websockets connections on all workers, it would be scalable without installing any additional tools.

miguelgrinberg:
> But why would I use an external message broker like Redis if I can (maybe) use uWSGI which already runs all of my code in production?

For your own application you can totally do what works for you. If this code works, then by all means use it.

I do not understand how you made it work, though, because as we agreed earlier, multiple uWSGI processes cannot communicate over this method.

> The question is rather why not?

Well, because it does not work. You may have found a particular use case in which you made it work, but it does not as a general solution. I honestly don't see how this can work at all, but as I said, if it works for you, I cannot prevent you from using it.

johaven:
> I do not understand how you made it work, though, because as we agreed earlier, multiple uWSGI processes cannot communicate over this method.

It is on this issue that we do not agree. As I said on my previous message (containing the logs), I had not identified that the listen was launched only on a single process. If the signal is registered at the initialization of the class, everything is well received by all workers.

Check with this fixed version.

miguelgrinberg:
So you are starting a single uWSGI master with 4 workers? That is not a supported use case, since the uWSGI load balancer does not support sticky sessions. Documentation: https://python-socketio.readthedocs.io/en/latest/server.html#scalability-notes. Without sticky sessions long polling is completely broken.

johaven:
I already read the whole doc, I promise :)
I use nginx in front:
```
upstream uwsgicluster {
        ip_hash;
        server unix:/var/www/app/run/socket_ws_0;
        server unix:/var/www/app/run/socket_ws_1;
        server unix:/var/www/app/run/socket_ws_2;
        server unix:/var/www/app/run/socket_ws_3;
    }
```
Each worker is dedicated to a socket.

I think I solved my problem with the last commit, I still have some tests to do.

johaven:
Ok that's it, everything is working properly.

In fact the listener part is useless in this case since it is the internal uwsgi signals which are used.

Ps: just a tips for uwsgi, to dedicate sockets to workers, you can do this:

```
socket = %(rundir)/socket_ws_0
socket = %(rundir)/socket_ws_1
socket = %(rundir)/socket_ws_2
socket = %(rundir)/socket_ws_3
map-socket = 0:1
map-socket = 1:2
map-socket = 2:3
map-socket = 4:4
```

miguelgrinberg:
I'm sorry, but I'm not able to accept these changes into the project.

The configuration in which this works is very unusual (I have never seen it used before). Also your pub/sub implementation is not robust, as it is subject to race conditions if the rate events are emitted is fast enough. Pub/sub is really hard to do well, the supported implements are all standard solutions to this, while I understand that this solution might work for your particular project, rolling your own pub/sub is not something I would like to encourage. Sorry.

johaven:
It is not because it has not been seen before that it is not a good solution.
This is a somewhat meaningless argument.

I agree with you on the fact that it is not yet mature enough (I will manage the concurrency), I do not intend to leave it like that.  I wanted to see this feature with you first.

This solution works for all those who use uwsgi, not for a specific project but actually you have to be a little interested in what the others offer.

I realize that you are a rather closed person, who lacks knowledge on the aspects of uwsgi.  And instead of helping me achieve this functionality, you rather buried me throughout this conversation.  And I don't really like it.

I'm going to spend my time, and keep this manager for myself.  This is not a problem.

You can close of course !

miguelgrinberg:
> It is not because it has not been seen before that it is not a good solution.
> This is a somewhat meaningless argument.

You are starting an uWSGI cluster with its own load balancer, then exposing the sockets for each individual worker so that you can bypass said load balancer and use nginx instead. 

We can set the discussion of whether this is a good or bad configuration aside. What you need to consider is that once this code is merged it is my responsibility to maintain it going forward. I can't really approve a change that I would not recommend to my users, and that I believe will increase my support work.

> I agree with you on the fact that it is not yet mature enough

Well, why didn't you mention it before then? You had me review your change as if it was ready then get defensive when I point out you have bugs. 

> This solution works for all those who use uwsgi

Not true. You have to use uWSGI configured with manually selected sockets and manually selected mapping of sockets to processes. You also need to put nginx in front to replace the uWSGI load balancing. That's the only case this solution works for, assuming you had no bugs in the implementation, which you do.

Sorry you feel this way.

johaven:
It is a dialogue of the deaf :)

> You are starting an uWSGI cluster **with its own load balancer,** then exposing the sockets for each individual worker so that you can bypass said load balancer and use nginx instead.

This is exactly what is described here: https://flask-socketio.readthedocs.io/en/latest/#using-nginx-as-a-websocket-reverse-proxy.
Yourself propose this solution: _The load balancer must be configured to forward all HTTP requests from a given client always to the same **worker**._

I read worker or server ? oh ! worker ! Yes but in your nginx configuration you load balance to servers ? ah ? And I think it was our first misunderstanding ...

I do not think this is a final solution to distribute the load to 4 independent servers or to advise, as you do, to run on a single worker.
And most people finally create http sockets with different ports on the same machine, that's what's going on in issues and stackoverflows.
There is a gap between these 2 solutions and the most optimal is to be able to use 1 worker per CPU, which is rather recommended to take advantage of the many CPUs of our servers.

Yes i assign sockets to workers, but only the traffic from /socketio is processed in this way.
I distribute all http traffic (api, static ...) across all workers to precisely take advantage of the "natural" load balancing of uWSGI.
I manage hundreds of instances of webservices in production, which I mostly coded and implemented (the devops is fantastic ... or not ....), I know a little what I do :)

As you said, this sockets assignment is only required for long polling, with websocket transport no need to do this.
You offer a great solution it's a shame not to go all the way ... my opinion ...

> Well, why didn't you mention it before then? ..

I had the misfortune to have done my tests from 2 ips addresses which were distributed on the same worker, I noticed it a little late.
And i realized that I was missing a brick of understanding on the listen part, but I expected more constructive thinking rather than "it doesn't work, put away your pencils, bye".

This solution covers 90% of uses as it is there.
Traders do not have fun placing stock market orders measured in nanoseconds via socketio. You have to be realistic, we are talking about applications for events that requires ms (chat, events etc ...).

And if you had read a little, there is as much chance of losing a signal coming from uWSGI as a message of Redis (even less in reality).

> uWSGI signals are managed with sockets, so they are fully reliable. When you send an uWSGI signal, you can be sure that it will be delivered. The signal table is shared by all workers (and protected against race conditions by a shared lock).

We can always improve, i think that I can easily set up an ack system on the queue part but i will see that on my side now.

In short, sorry to have been irritating on my previous message, we do not have the same point of view :)

I have to thank you for the time you spend working on this module.

johaven:
If you ever have issues related to this need.
I created a repository to host the final solution.

https://github.com/johaven/python-socketio-uwsgi-manager

PR #494, opened 2020-05-28
fix: fixed route path for tornado server

someApprentice:
In the documentation, in Tornado server section, the route path is specified as `r"/socketio.io/"`, when `r"/socket.io/"` must be.


miguelgrinberg:
Thanks!

PR #497, opened 2020-05-31
Update server.rst

StephenSJBrown:
Mentioned that rooms are created in the global default namespace and take an optional argument "namespace" to define the namespace the room is created in.


miguelgrinberg:
Thanks. There are a few other sections in which I have never written the "and this also works for namespaces" bit. In part it was intentional, because for 99% of the users namespaces are an unnecessary complication.

I'll review your suggestion and will probably reorganize the docs so that I can discuss the other functions that also have namespace support.

PR #635, opened 2021-02-10
Catch TimeoutError

bokolob:
Some network problems could cause TimeoutError, but it isn't handled. Moreover, the thread dies silently without any attempts to recover.

The bug can be also reproduced by setting the socket timeout option in Redis URL.

Probably there are other exceptions that should be retried as well, e.g BusyLoadingError.

### DIFF
--- a/socketio/redis_manager.py
+++ b/socketio/redis_manager.py
@@ -78,7 +78,7 @@ class RedisManager(PubSubManager):  # pragma: no cover
                 if not retry:
                     self._redis_connect()
                 return self.redis.publish(self.channel, pickle.dumps(data))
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
Any idea what's the equivalent error for `aioredis` so that both are fixed in parity?

bokolob:
No idea yet, I don't use async server :( From the first glance, it's unclear and needs some investigation.

bokolob:
By the way - It's very upsetting when thread dies, and the rest of the application continues to live.  I'd fix that -  register a callback to call if thread stop, or crash the whole app (in flask-socketio case)

Issue #617, opened 2021-01-22
Enhancement/bug fix


eserzomcity:
This borders between an enhancement and a bug fix --

When the [asyncio]redis_manager module is listening to subscribed to the pubsub pipe, if the connection fails it attempts to reconnect.  The reconnect uses exponential back off and caps at 60 seconds -- but the retry time never resets upon a successful connection.  Suggesting that the retry time reset back to 1. 

(in redis_manager.py)
  ```
      connect = False
        while True:
            try:
                if connect:
                    self._redis_connect()
                    self.pubsub.subscribe(self.channel)
                    retry_sleep = 1
                for message in self.pubsub.listen():
                    yield message


```

On a long running process, the exponential backoff may eventually hit 60 and stay pinned there -- causing the user to experience significant delays in receiving messages.

granted, there is an another underlying issue here (the redis server) - but I believe this would result in a better experience for the user of the API

miguelgrinberg:
Yes, this makes sense.

PR #636, opened 2021-02-10
Reset sleep timer upon reconnect

eserzomcity:
Reset the exponentially increasing timeout if a reconnect is achieved

https://github.com/miguelgrinberg/python-socketio/issues/617


miguelgrinberg:
Merged manually. I added the same logic to the asyncio redis and pika managers.

PR #692, opened 2021-05-26
Fixed the problem of AsyncAioPikaManager repeatedly creating a connec…

wangjianweiwei:


wangjianweiwei:
This is my first time to contribute code😄. give me some advice.

socketio/asyncio_aiopika_manager.py
```diff
@@ -46,30 +46,32 @@ def __init__(self, url='amqp://guest:guest@localhost:5672//',
...
+        AsyncPubSubManager.initialize(self)
 
-    async def _queue(self, channel, exchange):
-        queue = await channel.declare_queue(durable=False,
-                                            arguments={'x-expires': 300000})
-        await queue.bind(exchange)
-        return queue
+    def initialize(self):
+        loop = asyncio.get_event_loop()
+        loop.create_task(self._initialize())
```
miguelgrinberg, line 71: This is a really strange way to initialize. It is also bound to race conditions, since you also call _initialize() in the _listen() method.

socketio/asyncio_aiopika_manager.py
```diff
@@ -46,30 +46,32 @@ def __init__(self, url='amqp://guest:guest@localhost:5672//',
...
 
     async def _publish(self, data):
-        connection = await self._connection()
-        channel = await self._channel(connection)
-        exchange = await self._exchange(channel)
-        await exchange.publish(
+        await self.listener_exchange.publish(
```
miguelgrinberg, line 74: This will fail if the connection to the queue breaks and needs to be re-established. The Redis manage, which is the one that the vast majority of people use, has connection retries, maybe we want that here as well?

socketio/asyncio_aiopika_manager.py
```diff
@@ -80,16 +82,7 @@ async def _listen(self):
...
-                    self.listener_channel = await self._channel(
-                        self.listener_connection
-                    )
-                    await self.listener_channel.set_qos(prefetch_count=1)
-                    exchange = await self._exchange(self.listener_channel)
-                    self.listener_queue = await self._queue(
-                        self.listener_channel, exchange
-                    )
-                    retry_sleep = 1
+                    await self._initialize()
```
miguelgrinberg, line 85: This is problematic, because you have one more _initialize() call above. Both will set the same variables.

socketio/asyncio_aiopika_manager.py
wangjianweiwei: What do you suggest? I want to do this well.

miguelgrinberg:
Given the lack of response and the fact that this PR is now outdated, I will close it. Please open a new PR and address my feedback if you want me to add this.

Issue #409, opened 2020-01-16
Initial reconnect


adrianlubitz:
The socketio is not automatically trying to reconnect when there is no Internet connection on the initial connect. I have the following code snippet:
```python
import socketio
sio = socketio.Client()

@sio.event
def connect():
    print("I'm connected!")

@sio.event
def connect_error():
    print("The connection failed!")

@sio.event
def disconnect():
    print("I'm disconnected!")
    
sio.connect(MYSERVER)
```
It works perfectly if the Internet connection is established by the time of the connect. If it breaks down later it will reconnect, just as intended.

But if there is no Internet connection on the initial connect the following Exception is raised:

```
Traceback (most recent call last):
  File "pyTest.py", line 18, in <module>
    sio.connect(MYSERVER)
  File "/home/al/anaconda3/envs/socketioTest/lib/python3.6/site-packages/socketio/client.py", line 270, in connect
    six.raise_from(exceptions.ConnectionError(exc.args[0]), None)
  File "<string>", line 2, in raise_from
socketio.exceptions.ConnectionError: Connection refused by the server
```

The behavior should be that it tries to reconnect until there it can reach the server.

I worked my way around it by having the connect in a loop with try/except but I feel like this should be shipped with this great library.
Thanks ;)

miguelgrinberg:
Your description is accurate, the initial connection is not retried. This behavior was intentional, to me it does not make sense to enter a reconnect loop when there has never been a connection established before. Open to add an option to enable this if you have a use case that justifies it, but I don't see it as working that way by default.

adrianlubitz:
The problem I see here is if you have an unstable connection you never know when you have a connection and when you don't. Therefore it makes sense, that a connection is established whenever it is possible.
As far as I remember the js implementation does it like that and I just tested the java implementation and here it is just like that.
Just wanted to let you know, that when I was starting to use your implementation I was expecting this behavior and was confused by the way it is.
Besides that I was super happy how clean everything is structured :+1:

snorfalorpagus:
If the connection error was always the result of a configuration problem (e.g. the wrong URL has been given) I would agree that throwing the exception and not retrying makes sense. But it could be that the socketio server simply isn't available when the client starts up.

[...]

PR #789, opened 2021-10-04
Draft: Retry initial connection failures

snorfalorpagus:
This PR implements the suggestion in #409.

A new option has been added to the `Client.connect` method: `fail_fast`. The default value of `True` is no different to the existing behaviour. If set to `False`, if the call to connect fails it will be retried using the same logic as for a disconnect.

* I've only implemented this for the synchronous client for now. If it looks good I can port the changes to the async client also.
* I'm not entirely happy with the duplicate default values in `connect`/`_retry_connect` but I can't see an easy way around this.
* The `fail_fast` name is copied from a similar flag in the `aio-pika` library but could be something else.

### DIFF
--- a/src/socketio/client.py
+++ b/src/socketio/client.py
@@ -251,7 +251,7 @@ class Client(object):
 
     def connect(self, url, headers={}, auth=None, transports=None,
                 namespaces=None, socketio_path='socket.io', wait=True,
-                wait_timeout=1):
+                wait_timeout=1, fail_fast=True):
         """Connect to a Socket.IO server.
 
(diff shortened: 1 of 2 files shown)


miguelgrinberg:
Honestly I expected the change would be much smaller. What I imagined is that if `connect()` failed it would still return success after starting the reconnect thread in the background, so both the direct reconnect and the reconnect after a connection is lost would be handle in an identical way.

snorfalorpagus:
OK I see how that would work. I'll make the change.

MariusVB:
What happened to this? I'm looking for this functionality.

Issue #771, opened 2021-08-23
Cannot use AsyncRedis??


ahmetkca:
```
class MySocketManager:
	def __init__(self, app: FastAPI) -> None:
		self._mgr = socketio.AsyncRedisManager(REDIS_TLS_URL)
		self._sio: AsyncServer = socketio.AsyncServer(
			client_manager=self._mgr, 
			async_mode=ASYNC_MODE, 
			cors_allowed_origins='*'
		)
		self._app = socketio.ASGIApp(
			socketio_server=self._sio,
			socketio_path=SOCKETIO_PATH
		)
		app.mount(MOUNT_LOCATION, self._app)
		app.sio = self._sio

	def get_socket_manager(self) -> AsyncServer:
		return self._sio
```

and getting this  error
```
Traceback (most recent call last):
  File "/home/ahmetk/.local/share/virtualenvs/backend-pMCPp171/lib/python3.9/site-packages/socketio/asyncio_pubsub_manager.py", line 151, in _thread
    message = await self._listen()
  File "/home/ahmetk/.local/share/virtualenvs/backend-pMCPp171/lib/python3.9/site-packages/socketio/asyncio_redis_manager.py", line 93, in _listen
    self.sub = await aioredis.create_redis(
AttributeError: module 'aioredis' has no attribute 'create_redis'
```

miguelgrinberg:
It looks like just a couple of weeks ago aioredis released version 2, which is a complete rewrite and is not backwards compatible with version 1. For now, I suggest you downgrade your aioredis to the latest version 1. I'll keep this issue open to remind me to look at migrating to version 2.

ahmetkca:
thanks for the reply but i didn't even install aioredis it is not even in my Pipfile.

miguelgrinberg:
@ahmetkca I'm sorry to contradict you, but you do have aioredis installed and you have the new version, which was released within the last few weeks. The error that you are getting is not "you don't have aioredis", it is "something in your aioredis isn't what I expect". So check again, please.

Issue #505, opened 2020-06-11
[eventlet] how to socket.emit from the callback function


LaionFromNight:
Hi @miguelgrinberg  I hope you are fill ok

**Project description**
I'm currently building an application to visualization pressing key pianino
for handling midi adapter I'm using this lib rtmidi.

for practice, my python skill I pickup to create a midi-adapter on python and via socket connect it into an angular application.

**The my issue**
the PoC works fine, but I got a big delay in emitting so I was star searching and I found some example how to implement the eventlet and on this key testing via using
setup_midi_port_mock and setup_midi_port_mock_5_keys looks pretty good but emit from callback stop working ;(

**POC project**
as I mentioned before I got a big delay | this screnshot is for this setup on def init:
setup_midi_port_mock_5_keys

**left time** is from the client after receive **right side** is from python server
![image](https://user-images.githubusercontent.com/66757837/84333162-38cf3c80-ab8f-11ea-8802-9400f5337dad.png)

```
[[source]]
name = "pypi"
url = "https://pypi.org/simple"
verify_ssl = true

[dev-packages]

[packages]
python-rtmidi = "*"
flask = "*"
flask-socketio = "*"

[requires]
python_version = "3.6"
```

```
from flask import Flask
from flask_socketio import SocketIO
import rtmidi
from datetime import datetime
'''
 Setup Socket IO
'''

app = Flask(__name__)
socket = SocketIO(app, logger=True, engineio_logger=True, cors_allowed_origins="*")
# app.debug = True
# app.host = '0.0.0.0'

'''
    Setup midiIn
'''
midiIn = rtmidi.MidiIn()




@socket.on('init')
def init(data):
    setup_midi_port()
    # setup_midi_port_mock()
    # setup_midi_port_mock_5_keys()


def rise_key(data):
    emit_data = {
        'action': data[0][0],
        'key': data[0][1],
        'emphasis': data[0][2],
        'duration': data[1],
        'time': datetime.utcnow().strftime('%Y-%m-%d %H:%M:%S.%f')[:-3]
    }
    socket.emit('rise-key', emit_data, broadcast=True, namespace='/')
    socket.sleep(0.0001)


def click_key(data):
    emit_data = {
        'action': data[0][0],
        'key': data[0][1],
        'emphasis': data[0][2],
        'duration': data[1],
        'time': datetime.utcnow().strftime('%Y-%m-%d %H:%M:%S.%f')[:-3]
    }
[...]

PR #801, opened 2021-10-16
Migrate to aioredis 2.0

sam-mosleh:
This PR fixes issue #771

### DIFF
--- a/src/socketio/asyncio_aiopika_manager.py
+++ b/src/socketio/asyncio_aiopika_manager.py
@@ -94,7 +94,7 @@ class AsyncAioPikaManager(AsyncPubSubManager):  # pragma: no cover
                 async with self.listener_queue.iterator() as queue_iter:
                     async for message in queue_iter:
                         with message.process():
-                            return pickle.loads(message.body)
+                            yield pickle.loads(message.body)
             except Exception:
                 self._get_logger().error('Cannot receive from rabbitmq... '
                                          'retrying in '
--- a/src/socketio/asyncio_pubsub_manager.py
+++ b/src/socketio/asyncio_pubsub_manager.py
@@ -148,35 +148,35 @@ class AsyncPubSubManager(AsyncManager):
     async def _thread(self):
         while True:
             try:
-                message = await self._listen()
+                async for message in self._listen():
+                    data = None
+                    if isinstance(message, dict):
+                        data = message
+                    else:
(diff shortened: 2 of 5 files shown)


miguelgrinberg:
You have changed a lot of other things besides upgrade aioredis. Why?

miguelgrinberg:
Thanks. After looking at your PR in more detail I do see that these changes are justified. I have now merged your PR with some changes which made the code closer to the regular Redis client manager. Thank you!

miguelgrinberg:
@sam-mosleh To me the most important thing is to have parity between the two Redis implementations. I did not really look at every minor change you've made, I just replaced your solution with the existing one in the sync redis. If you have any suggestions to improve the redis implementation you are welcome to submit a PR targeting both, which are now almost identical.

miguelgrinberg:
@cnicodeme I was hoping to look into https://github.com/miguelgrinberg/python-socketio/pull/806 more in detail before releasing, but haven't had time to look into that yet.

miguelgrinberg:
@cnicodeme 5.5.0 is now out with this change.

cnicodeme:
Awesome, thank you @miguelgrinberg !

cnicodeme:
I just updated to 5.5.0 and noticed a previous bug that is still present (Also mentioned in #505). 

The problem is indeed a backward incompatible update from Sanic : They moved the HTTPResponse and WebSocketProtocol to another path, making the import to fail.

I've previously brought a suggestion at: https://github.com/miguelgrinberg/python-engineio/issues/251

miguelgrinberg:
@cnicodeme I consider Sanic support much lower priority, did not plan to have it fixed in this release. In fact I've been thinking about dropping Sanic from the support list, because they often break things, it's a lot of effort to keep up with them. The Sanic implementation is already hacky because they make it extremely difficult to support a route that is both HTTP and WebSocket.

cnicodeme:
Yeah I can understand the issue.

I'm all ok with dropping the support, but in that case, would there be - if not already - a way to implement a custom integrations?

I'm mentioning this because in the current structure, it's not possible to extract the Sanic integration and use `async_mode` when creating the AsyncServer, because the mode 'sanic' is hard written in the `AsyncServer` class at https://github.com/miguelgrinberg/python-engineio/blob/main/src/engineio/asyncio_server.py#L76

PR #806, opened 2021-10-20
Improve redis usage

sam-mosleh:
This PR improves Redis usage by avoiding creating a connection pool for every connection failure which is not needed for both redis and aioredis libraries.


miguelgrinberg:
I have incorporated the `ignore_subscribe_messages` option, but I'm not going to include the rest of this PR, mainly because the current implementation has been used for a long time and works reliably, and I do not have the time to retest and ensure that your solution is equally robust (my impression is that it isn't, actually). Thanks.

sam-mosleh:
It makes sense.

PR #844, opened 2022-01-10
added django_example for asgi servers with uvicorn

Psami-wondah:


miguelgrinberg:
Unfortunately I do not believe this to be a good example for people to model their applications after. The main problem I see is that you have put the entire Socket.IO application in asgi.py. Other minor problems are that you have references to an app called `socketio` in the settings that does not exist and you have the database binary file committed to the repository. It's also confusing that you have two methods for starting the server in the same repo (Uvicorn and Gunicorn).

Psami-wondah:
The gunicorn is set up for deployment. But locally uvicorn is used to run the app. As for the references to the socketio app I'll fix that. I previously named an app socketio and it conflicted with the socketio package so I deleted it. Then I'll work on implementing the sio in views and update my PR.

Psami-wondah:
@miguelgrinberg I have made updates to my PR. You can check it again now. I also unstaged the database binary file.

miguelgrinberg:
@Psami-wondah all the server examples have the same functionality regardless of the framework, designed for people to test pretty much all the options available. Why did you decide to implement a completely different application for this that uses less of the library?

Psami-wondah:
I use django a lot and I always thought it wasn't possible to use  socketio on it. After discovering this library I decided to add an example so others can see the possibility of the functionality. I've always used django channels for web sockets. But socketio has more functionality than websockets. So I'll be using this from now onwards.

miguelgrinberg:
Yeah, the question was why this application is different than all the others? I have implemented the same application for every single framework. This one is different.

Psami-wondah:
It's not different. There's just no Django_example for the asgi servers in the repo.

miguelgrinberg:
@Psami-wondah you wrote a different application, don't tell me it isn't different. I have ported the exact same application to all the frameworks, including Django WSGI. Instead of porting the application to Django ASGI you wrote a different application, which continues to be less ideal and not a good model in some aspects compared to the demo application that I built.

So what I'm going to do is reject this PR and when I get a moment I'll port the WSGI app to ASGI.

Psami-wondah:
I think I understand you now. So I should have gotten the wsgi django app and ported it to ASGI instead of starting a new django app.

miguelgrinberg:
Closing, per discussion above.

sh1man999:
2 years have already passed when you can see the result of the work of Djnango + asgi

PR #850, opened 2022-01-15
PubSub: Allow custom encoder

Ododo:
Hello.

I wanted to introduce this functionality because i needed the RedisManager to use Json instead of pickle for encoding data, so that other non-python applications listening to the redis channel can decode the data more easily.

I extended this to allow any encoder object to be used basically.

Thanks,

Olivier

### DIFF
--- a/src/socketio/redis_manager.py
+++ b/src/socketio/redis_manager.py
@@ -36,21 +36,26 @@ class RedisManager(PubSubManager):  # pragma: no cover
                        and receiving.
     :param redis_options: additional keyword arguments to be passed to
                           ``Redis.from_url()``.
+    :param encoder: The encoder to use for publishing and decoding data,
+                    defaults to pickle.
     """
     name = 'redis'
 
     def __init__(self, url='redis://localhost:6379/0', channel='socketio',
-                 write_only=False, logger=None, redis_options=None):
+                 write_only=False, logger=None, redis_options=None,
+                 encoder=pickle):
         if redis is None:
             raise RuntimeError('Redis package is not installed '
                                '(Run "pip install redis" in your '
                                'virtualenv).')
+        self.encoder = encoder
         self.redis_url = url
         self.redis_options = redis_options or {}
         self._redis_connect()
         super(RedisManager, self).__init__(channel=channel,
                                            write_only=write_only,
-                                           logger=logger)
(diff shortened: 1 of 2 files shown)


miguelgrinberg:
As far as I can see this PR is incomplete. You are decoding messages using the provided encoder, but you are not encoding with it.

Ododo:
So there are 2 commits, one In the PubSubManager for decoding and the other for encoding in the RedisManager only (i tested only this).
Do you want me to add encoder parameters for all the manager classes ?

src/socketio/redis_manager.py
```diff
@@ -36,21 +36,26 @@ class RedisManager(PubSubManager):  # pragma: no cover
...
 
     def __init__(self, url='redis://localhost:6379/0', channel='socketio',
-                 write_only=False, logger=None, redis_options=None):
+                 write_only=False, logger=None, redis_options=None,
+                 encoder=pickle):
         if redis is None:
             raise RuntimeError('Redis package is not installed '
                                '(Run "pip install redis" in your '
                                'virtualenv).')
+        self.encoder = encoder
```
miguelgrinberg, line 51: This isn't necessary, right? The parent class also stores this attribute.

miguelgrinberg:
> So there are 2 commits, one In the PubSubManager for decoding and the other for encoding in the RedisManager only (i tested only this).

Sorry, yes, I missed the RedisManager change.

> Do you want me to add encoder parameters for all the manager classes ?

Yes, I want to have parity across these classes.

The build results also show that now there is a portion of the code that isn't covered by tests. Would you be able to address that as well?

src/socketio/redis_manager.py
Ododo: right, thanks

Ododo:
> The build results also show that now there is a portion of the code that isn't covered by tests. Would you be able to address that as well?

Yes, will do.

Ododo:
Hi @miguelgrinberg
I wanted to let you know that this PR is ready to be reviewed.

Kr

DEATHSOULLQ8 (approved):

miguelgrinberg:
Moved to https://code.miguelgrinberg.com/miguelgrinberg/python-socketio/issues/10004

Issue #822, opened 2021-11-29
Client is always accepted for undeclared namespaces


llunn:
**Describe the bug**
This issue is manifesting itself through flask-socketio, but I believe the source of the issue belongs here. When a client connects using a namespace that is not declared, the namespace is simply added and the client connects with no ability to execute code against the connection in order to reject it.  

The end result is that it is possible to pollute the connection pool with phantom clients, and the server cannot detect when new clients connect. 

The source of issue I believe is here:  https://github.com/miguelgrinberg/python-socketio/blob/3bd13578c82e8a94d6b0328180606c2aefd496f1/src/socketio/base_manager.py#L119

**To Reproduce**

1. Stand up a bare bones server instance.
2. Define an on_connect handler for '/'
2. Define zero or more class based namespaces and register them
3. Connect a client using a namespace that the server is not aware of.
4. Client connection is blindly accepted, regardless of `always_connect` value.

**Expected behavior**
One of:

1. Connections to namespaces that do not exist are flat out rejected; or
1. Connections to namespaces that do not exist produce an exception; or
2. Connect event is delegated to to `/` namespace `connect`, with relevant details on the namespace attempted to connect to.

In all cases, it is also expected that the `/` namespace `connect` event fires, which is also not occurring.

**Logs**
```
(58685) wsgi starting up on http://0.0.0.0:5000
(58685) accepted ('127.0.0.1', 50507)
1_szfyPlEneUJwA3AAAA: Sending packet OPEN data {'sid': '1_szfyPlEneUJwA3AAAA', 'upgrades': [], 'pingTimeout': 20000, 'pingInterval': 25000}
kombu backend initialized.
1_szfyPlEneUJwA3AAAA: Received request to upgrade to websocket
1_szfyPlEneUJwA3AAAA: Upgrade to websocket successful
1_szfyPlEneUJwA3AAAA: Received packet MESSAGE data 0/celery,{"authToken":"bGVlQHRvcnVzb2Z0LmNvbTo2MUE1MUI1NTrMnz4SGR56ssXsYgxiiGHGa5ck6g"}
/celery
1_szfyPlEneUJwA3AAAA: Sending packet MESSAGE data 0/celery,{"sid":"hOsZtTCikRL7RTIfAAAB"}
(58685) accepted ('192.168.0.60', 50513)
[...]

PR #856, opened 2022-01-19
Fix connecting to undeclared namespaces

sam-mosleh:
This PR fixes issue #822

Accepting client connections for an unknown namespace could be problematic and hard to debug besides security issues.
In this PR new exception (**NamespaceNotFoundError**) is introduced which will prevent clients to connect with reason.


miguelgrinberg:
Thanks. I used your code as input, but I ended up addressing this problem with a simpler fix.

PR #872, opened 2022-02-14
 Fix indentation in the "Rooms" docs example.

ezio-melotti:
[The first example of the "Rooms" section of the Socket.IO Server docs](https://python-socketio.readthedocs.io/en/latest/server.html#rooms) is not indented properly.  This PR fixes the indentation.


miguelgrinberg:
Thanks!

PR #892, opened 2022-03-16
Update app.py

flocko-motion:
Sanic applications require a name. Otherwise error: sanic.exceptions.SanicException: Sanic instance cannot be unnamed. Please use Sanic(name='your_application_name') instead.


miguelgrinberg:
Thanks!

PR #935, opened 2022-06-16
Try to reuse already created connection and channel while emitting from external processes

theDigitalGuy:
to avoid rabbitmq starting to block connections.

Problem description: https://github.com/theDigitalGuy/python-socketio/commit/4ff70fb06c0c5b407f745ce0755a7692888de5c5#commitcomment-76278910

### DIFF
--- a/src/socketio/asyncio_aiopika_manager.py
+++ b/src/socketio/asyncio_aiopika_manager.py
@@ -55,7 +55,7 @@ class AsyncAioPikaManager(AsyncPubSubManager):  # pragma: no cover
         super().__init__(channel=channel, write_only=write_only, logger=logger)
 
     async def _connection(self):
-        if (self.aiopika_connection is None or self.aiopika_connection.is_closed is False):
+        if (self.aiopika_connection is None or self.aiopika_connection.is_closed is True):
             return await aio_pika.connect_robust(self.url)
         return self.aiopika_connection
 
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
There's two problems:

1. Your solution is not robust enough. It should handle random disconnects and retries. See the Redis implementations to have an idea of what is expected.
2. The Kombu and aiopika implementations are more or less the same. If one is improved, the other should be improved in the same way.

sillyfrog:
Regarding (1), the `RobustConnection` object returned by `connect_robust` actively reconnects on errors/disconnects etc. I've done extensive testing in single node and clustered configurations to confirm it's very robust.

While there is no connection to the server an error is raised, but as soon as connectivity is restored, things start working again as you would hope, (probably) negating the need for this to be re-implemented in `python-socketio`.

src/socketio/asyncio_aiopika_manager.py
```diff
@@ -47,17 +47,28 @@ def __init__(self, url='amqp://guest:guest@localhost:5672//',
...
+        self.aiopika_channel = None
+        self.aiopika_exchange = None
+
         super().__init__(channel=channel, write_only=write_only, logger=logger)
 
     async def _connection(self):
-        return await aio_pika.connect_robust(self.url)
+        if (self.aiopika_connection is None or self.aiopika_connection.is_closed is True):
+            return await aio_pika.connect_robust(self.url)
+        return self.aiopika_connection
```
sillyfrog, line 60: This should have a lock around it as if there are several calls in close succession, it'll create several new connections. See #1142 for an example of this in use.

miguelgrinberg:
Fixed by https://github.com/miguelgrinberg/python-socketio/commit/cd7f781c022dd1d1ec3c6695a0fd6ab3ce864fd5.

PR #941, opened 2022-06-23
Fix deprecation warning

jboulmier:
In `emit` method of `asyncio_manager`, coroutines objects are directly passed to `asyncio.wait`, this comportment is deprecated since python 3.8 and will be removed in python 3.11, see [python doc](https://docs.python.org/3/library/asyncio-task.html#waiting-primitives) for more information.

https://github.com/miguelgrinberg/python-socketio/blob/5b9134617759a1b64adb2f9aba0974c732576cc4/src/socketio/asyncio_manager.py#L11-L33

This patch fixes that problem by adding the required `asyncio.create_task`

### DIFF
--- a/src/socketio/asyncio_manager.py
+++ b/src/socketio/asyncio_manager.py
@@ -26,8 +26,9 @@ class AsyncManager(BaseManager):
                     id = self._generate_ack_id(sid, callback)
                 else:
                     id = None
-                tasks.append(self.server._emit_internal(eio_sid, event, data,
-                                                        namespace, id))
+                tasks.append(asyncio.create_task(
+                    self.server._emit_internal(eio_sid, event, data,
+                                               namespace, id)))
         if tasks == []:  # pragma: no cover
             return
         await asyncio.wait(tasks)


miguelgrinberg:
Looks good, thanks. Was this the only occurrence? I'm surprised I don't have more of these to update, either here or in the engineio package.

rodolfosouza:
guys I tried to update from 5.6 to 5.7 and this change is not compatible with python 3.6 anymore
ERROR:  module 'asyncio' has no attribute 'create_task'"

https://github.com/miguelgrinberg/python-socketio/commit/4b697815c3da4574fca8b759d21a4e0800dafc50

jboulmier:
I haven't handled this case since support for python3.6 seems to have been removed in this commit:
https://github.com/miguelgrinberg/python-socketio/commit/ed5679a7cb01963b44a5004e1236b7e8b485aa0b

miguelgrinberg:
@rodolfosouza 3.6 isn't in the list of supported versions anymore.

PR #966, opened 2022-07-17
Fix error typo: redis -> aioredis

zt50tz:
`AsyncRedisManager` module provide error message with `redis` package instead of `aioredis` if it is not installed. This might be embarrassing for someone.

Thanks for your awesome package!


miguelgrinberg:
This is actually correct. The `redis` package supports asyncio now.

zt50tz:
Seems my installed version was too old. Thanks.

PR #979, opened 2022-07-23
get number of clients in a room

ebottos94:
hi @miguelgrinberg I've created two function in server.py and base_manager.py in order to have the number of clients in a room.


miguelgrinberg:
This has been discussed a lot of times. Returning the number of people in a room is actually quite involving, and I have long ago decided to leave this task to applications, who are better able to maintain room counts.

Your PR breaks for the case when multiple servers are used.

PR #981, opened 2022-07-27
Update msgpack_packet.py

Cromfel:
Based on the official documentation payload is optional. 

Problem: This caused issues when using python-socketio together with other implementations.

Currently:
`self.data = decoded['data']`

Should be:
`self.data = decoded.get('data')`

Reference:
https://github.com/miguelgrinberg/python-socketio/blob/d556a0ec4de5ef2ca21626ce8b41d68126ec3605/src/socketio/msgpack_packet.py#L16

Documentation: 
https://github.com/socketio/socket.io-protocol#packet-format


miguelgrinberg:
Thanks!

Issue #390, opened 2019-11-29
How use socketio decorator within another class


sraboncmc:
Hi Miguel
I tried to use socketio.Client within a class where i can call socketio and its related event frequently.But dnt understand how figure out it.
It works fine on normal python file.

i m a part time coder,so plz dnt laugh on my coding.
Thanks in advance.
```python
import socketio

class warSock:

    global mwar
    global name
    global ps
    mwar = socketio.Client()

    def __init__(self, name, ps):
        self.name = name
        self.ps = ps

    @mwar.event
    def connect(self):
        print('connected to server')

    @mwar.on('login')
    def on_login(data):
        print data

    def connectme(self):
        mwar.connect('http://localhost')

    def dc(self):
        mwar.disconnect()
```

miguelgrinberg:
What part of this does not work? The only thing I noticed is that the on_login method is missing `self`.

As a side note, you may want to look at class-based namespaces in the documentation, as they already provide a way to implement events as methods in a class.

sraboncmc:
Thanks for your reply.
Problem was from here

> @mwar.event
    def connect(self):
        print('connected to server')

sio wants connect(), but here class wants connect(self)
thats why sio call argument typeError (something like this)

sraboncmc:
And i think i sort out this by putting all sio related events inner function(def) of class def. And it works still.
And So much thanks for your great creation.I tried it 9 month earlier,bt i failed to connect single one.
may be i failed to understand the concept of decorator.

I have a request for connection establishment. It got hang on ide as like httplib connection.
I used socketio js and java client.Those are faster.

Zartris:
This might be old and you have solved it.
If you don't need the self functionality in connect just make it static like this:
    
@staticmethod
@mwar.event
def connect():
     print('connection established')

Robokishan:
i know this is old but i wanted to have this kind of functionality and i have done this

# # socketIO.wait(seconds=1)
```
import socketio
import json

class Wrapper_class():
        sio = socketio.Client()

        def __init__(self):
            pass
        def setup(self):
            self.call_backs()
            self.sio.connect('http://localhost.local:8081')

        def loop(self): 
            self.sio.wait()

        def call_backs(self):
            @self.sio.event
            def connect():
                self.sio.emit('subscribe','room')
                print('connection established')
                
            @self.sio.on("docs")
            def raw_data(data):
                print(f"Data Received {data}")


            @self.sio.event
            def auth(data):
[...]

PR #1056, opened 2022-10-06
(seatedscribe) Let reconnect use wait parameters

seatedscribe:
If a client is successfully connected to a server which is later pulled down for a few seconds, what happens at the reconnection stage? 
In my case the connected flag needed to be true as soon as the 'connect' event is thrown, not after one second (default implicit value)

So I propose this PR to put coherence between connect and reconnect algos, hope it is helpful!

### DIFF
--- a/src/socketio/asyncio_client.py
+++ b/src/socketio/asyncio_client.py
@@ -122,6 +122,8 @@ class AsyncClient(client.Client):
         self.connection_transports = transports
         self.connection_namespaces = namespaces
         self.socketio_path = socketio_path
+        self.wait = wait
+        self.wait_timeout = wait_timeout
 
         if namespaces is None:
             namespaces = list(set(self.handlers.keys()).union(
@@ -479,7 +481,9 @@ class AsyncClient(client.Client):
                                    auth=self.connection_auth,
                                    transports=self.connection_transports,
                                    namespaces=self.connection_namespaces,
-                                   socketio_path=self.socketio_path)
+                                   socketio_path=self.socketio_path,
+                                   wait=self.wait,
+                                   wait_timeout = self.wait_timeout)
             except (exceptions.ConnectionError, ValueError):
                 pass
             else:
--- a/src/socketio/client.py
+++ b/src/socketio/client.py
@@ -311,6 +311,8 @@ class Client(object):
         self.connection_transports = transports
         self.connection_namespaces = namespaces
         self.socketio_path = socketio_path
+        self.wait = wait
+        self.wait_timeout = wait_timeout
 
         if namespaces is None:
             namespaces = list(set(self.handlers.keys()).union(
@@ -662,7 +664,9 @@ class Client(object):
                              auth=self.connection_auth,
                              transports=self.connection_transports,
                              namespaces=self.connection_namespaces,
-                             socketio_path=self.socketio_path)
+                             socketio_path=self.socketio_path,
+                             wait=self.wait,
+                             wait_timeout=self.wait_timeout)
             except (exceptions.ConnectionError, ValueError):
                 pass
             else:


seatedscribe:
```

import socketio

from src.backend import Backend

sio = socketio.Client()  # logger=True, engineio_logger=True)
backend = Backend()

def pool_update(card_info: dict):
    print(f'pool update - connected: {sio.connected}')
    sio.emit('pool_update', card_info)
    

@sio.event
def connect():
    pool_update(backend.get_status())
    [...]

while not sio.connected:
    try:
        sio.connect(url='http://127.0.0.1:5000', wait=False)
    except Exception:
        print("Trying to connect...")
        sio.sleep(2)
```

I had an exception inside connect(), which I am currently not able to reproduce
Strange enough, I am not currently able to reproduce the issue, probably unrelated with Client.connected flag.
Sorry for bothering you, but trying to avoid the exception made me adding controls

miguelgrinberg:
I'm not sure what was the original issue you had, so it would be good if you can figure out what that was, and then we can go from there. As things stand, I don't really see a value in passing the wait options during reconnect.

seatedscribe:
I am currently able to reproduce the issue. 👀 
The above snippet has been edited to reflect current situation and can be used as reference.
The exception raised is a BadNamespaceError triggered by `sio.emit('pool_update', card_info)` in the function `def pool_update(card_info: dict):` (traceback follows)

--- context ---
I have a smartcard reader and an observer which calls the pool_update() upon insertion/removal of cards who is responsible to emit the updated data to the server (on the same machine for testing purposes).

As long as the _sio server_ is up and running, everything is good.

If I **shut down** the server, as soon as pool_update() is called (card is removed/inserted) the exception is raised. Strange enough, this happens today out of the blue, with basically no changes in the codebase. You can still see the `sio connected? False` log message, which seemed to me the root of the problem and hence the reason for my patch.

This is the traceback:
```
[2022-10-13 10:55:06,707 INFO frontend.py:43 -           disconnect()] disconnected from server
[2022-10-13 10:55:14,146 WARNING watchers.py:59 -               update()] Removed card XXX804, slot 0
[2022-10-13 10:55:14,149 INFO cardcomm.py:86 -             finalize()] slot 0, finalize reply 0, True
[2022-10-13 10:55:14,151 WARNING backend.py:111 -       _teardown_card()] Removed sn C0137, slot 0
[2022-10-13 10:55:14,152 WARNING frontend.py:20 -          pool_update()] **pool update - sio connected? False**
Exception in thread Thread-1:
  File "/usr/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/opt/siae-app_MS1482/pool/venv/lib/python3.7/site-packages/smartcard/CardMonitoring.py", line 179, in run
    (addedcards, removedcards))
  File "/opt/siae-app_MS1482/pool/venv/lib/python3.7/site-packages/smartcard/Observer.py", line 62, in notifyObservers
    observer.update(self, arg)
  File "/opt/siae-app_MS1482/pool/src/watchers.py", line 60, in update
    self._callback(CardEvent(slot, atr, Direction.OUT))
  File "/opt/siae-app_MS1482/pool/src/backend.py", line 47, in handle_card_event
    self._teardown_card(slot=event.slot, atr=event.atr)
  File "/opt/siae-app_MS1482/pool/src/backend.py", line 114, in _teardown_card
    self.exec_status_callback()
  File "/opt/siae-app_MS1482/pool/src/backend.py", line 130, in exec_status_callback
    self.client_callback(self.get_status())
  File "/opt/siae-app_MS1482/pool/src/frontend.py", line 21, in pool_update
    sio.emit('pool_update', card_info)
  File "/opt/siae-app_MS1482/pool/venv/lib/python3.7/site-packages/socketio/client.py", line 394, in emit
    namespace + ' is not a connected namespace.')
socketio.exceptions.BadNamespaceError: / is not a connected namespace.
```

This is the venv:
```
aiohttp==3.8.3
aiosignal==1.2.0
async-timeout==4.0.2
asynctest==0.13.0
attrs==22.1.0
backports.zoneinfo==0.2.1
bidict==0.22.0
certifi==2022.9.24
cffi==1.15.1
charset-normalizer==2.1.1
coverage==6.5.0
cryptography==3.4.7
frozenlist==1.3.1
idna==3.4
multidict==6.0.2
netifaces==0.11.0
nop==0.1.1
pkg_resources==0.0.0
pycparser==2.21
pyscard==2.0.5
python-engineio==4.3.4
python-socketio==5.7.1
pytz==2022.4
pytz-deprecation-shim==0.1.0.post0
requests==2.28.1
typing_extensions==4.4.0
tzdata==2022.4
tzlocal==4.2
urllib3==1.26.12
websocket-client==1.4.1
yarl==1.8.1
```

Thanks in advance

miguelgrinberg:
I'm sorry, but there is something I don't understand. You printed `sio.connected` and it was `False`, correct? So why would you then try to emit on this obviously disconnected client? And how is this related to the reconnection process? It seems regardless of what the wait flags are, if you call this `pool_update()` function at the wrong moment you are always going to get the exception, because you emit without checking if the client is connected or not.

seatedscribe:
Exactly, this is the same thought I had! 🎉
If you see in `pool_update()` I had to add (currently commented) code and skip emitting if connection is not active:
```
def pool_update(card_info: dict):
    _LOGGER.warning(f'pool update - sio connected? {sio.connected}')
    sio.emit('pool_update', card_info)
    if sio.connected:
        _LOGGER.info("emitting pool update")
        sio.emit('pool_update', card_info)
    else:
        _LOGGER.warning("NOT emitting pool update - yet disconnected")
```

Still, looking around the web I don't see much code that handles the connection state, hence I have a doubt.

> 1. Is checking for `sio.connected` the right way? What would your approach be in such situations where `emit()` is triggered by external hardware (backend), regardless of network health? `backend.set_status_callback(pool_update)`


Because imho adding "low-level" checks in application code wheighs down the end user a bit
> 2. Shouldn't `Client/emit` take care of connection state internally (e.g. Client can be configured to silently discard the message or explicitly raise an exception) relieving end users the chore to check for connection flags (and their timeouts)?

> 3. What is the reason a BadNamespaceError thrown? I would rather expect a ConnectionError in this case.

Back to my story (happy ending):
Since now in `pool_update()` I have to check for `sio.connected` every time before emitting, I turned off wait feature `sio.connect(..., wait=False)` so the flag is True instantly at startup AND (I thought) at **reconnection** stage.

The outcome now is that `emit()` is executed on first connection, but not _on each subsequent reconnection_, the reason being that flag is false (for one second), in turn because `_handle_reconnect()` does not honor original `wait=False` parameter. The funny thing is that `emit()` would succeed if actually executed, as connection is ok - client has just been notified that the server is up again!)

I hope it is clearer now why this patch is important. I agree `_handle_reconnect()` is a background task _per-se_, but has side effects if implementation details have to bubble up to end user code, like sio.connected precisely. And this bring us back to question 1. 😄

miguelgrinberg:
> Is checking for sio.connected the right way? 

No. The correct approach is to have connect and disconnect handlers, and update your application state in those. There is really no need to use the internal `connected` flag in your application, since you are always told when the connection goes up or down through the handlers.

A Socket.IO connection cannot simply be "connected" or "disconnected", as there are at least two layers (sometimes more) and these are not always in the same connected state. The Engine.IO layer connects first, then each namespace is connected on top. As long as you have connect/disconnect handlers on the intended namespace, you are always going to be told when that namespace is connected or not.

> Shouldn't Client/emit take care of connection state internally

In the current design the connection up/down is passed to the application. You have the option to maintain your own connection state through the connect/disconnect handlers and then check before you emit, or you can also just emit and catch connection and bad namespace errors, which is more Pythonic.

> What is the reason a BadNamespaceError thrown?

Going back to the above, this happens because the connection is established in the low level, but the namespace that you are using hasn't been connected yet.

> I hope it is clearer now why this patch is important.

Unfortunately not. I don't really follow your explanation of what happens, but I have the following comments to make:

- Passing `wait=False` when you connect rarely makes sense. I don't understand why this makes sense in your case.
- As stated above, the `sio.connected` flag is internal, and is unable to express all the complexities involved in a Socket.IO connection. So you shouldn't base any decisions made in your application on this flag.

seatedscribe:
I am confused.
Let split the issues: I am actually using connect/disconnect handlers, as you already saw in the snippet.

```
@sio.event
def connect():
    _LOGGER.info('server connected')
    pool_update(backend.get_status())
```

Based on what you said the exception should not happen then, or this is not the right handle for re/connections?
Inside the handler I call a function `pool_update` that `emit`s and there an exception is raised.

seatedscribe:
Ok, I understand the domain is different, so there is a way to structure a class that inglobes the handlers, instead defining functions at the module level, would this be possible? I tried without success until now

miguelgrinberg:
@seatedscribe Can you use inner functions? or manually invoke the decorator as a function, so that you can decorate your handlers after your `SocketIO` object is created?

seatedscribe:
I finally found a viable solution [here](https://github.com/miguelgrinberg/python-socketio/issues/390#issuecomment-787796871) with the addition of a local variable to take care of connected state wrt the external hardware triggering the `emit()`.

> You have made a reference once again to a one second delay that happens during reconnection. Still don't know exactly what you refer to here.

Even if it is not important any more, just for completeness the timeout I am referring is this: 
https://github.com/miguelgrinberg/python-socketio/blob/2c3e360ae8b151bc0bfedbde50248cd0dc8d1ff9/src/socketio/client.py#L262-L263
which prevents sio.connected to be True instantly at reconnection, but only after 1 second, because of `wait` parameter not honored here:
https://github.com/miguelgrinberg/python-socketio/blob/2c3e360ae8b151bc0bfedbde50248cd0dc8d1ff9/src/socketio/client.py#L660-L665

I know, I know, not the right way.
Thanks anyway for time and effort 😄

miguelgrinberg:
> Even if it is not important any more, just for completeness the timeout I am referring is this:

Ah, okay. This is a timeout, not a delay. In most cases the connect function will return sooner than one second. In fact, if it takes a second, that means that the timeout happened before the connection was fully established, which is unusual.

PR #1103, opened 2022-12-22
feat: pass manager kwargs from namespace

devamin:
# WHY
I need to pass `ignore_queue` from namespace emit call. 
# WHAT 
Pass kwargs from namespace emit to server emit

### DIFF
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -577,6 +577,16 @@ The RabbitMQ queue is configured through the
     mgr = socketio.AsyncAioPikaManager('amqp://')
     sio = socketio.AsyncServer(client_manager=mgr)
 
+Ignore the queue
+~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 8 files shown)


miguelgrinberg:
Thanks, this looks good in principle, but passing `kwargs` is not a good solution, as it makes it much harder to properly document these calls. So I would like to request two things:

1. Add any missing function arguments explicitly
2. Update the docstrings accordingly

Do you think you can make these changes? Thanks!

devamin:
You welcome, thanks for you for creating this library. 
Yes I can handle these requirements.

devamin:
@miguelgrinberg please let me know if all good, or there is something that I need to change/add.

miguelgrinberg:
Thanks. There were a few missing things in this PR such as the same support for `ignore_queue` in `send()` and `call()`, and also the tests needed some updates to cope with these changes. I have committed a complete fix now.

PR #1108, opened 2022-12-31
ambiguous

iamsabbirsobhani:
Ambiguous between `wsgi` app and socket `server` instance `app`


miguelgrinberg:
This is intended. The name of a WSGI application instance is likely either `app` or `application`. Changing it to `wsgi_app` makes the example more confusing in my opinion.

The `WSGIApp` class acts as an application wrapper, so it produces a wrapped `app` using the original `app` as input.

Issue #691, opened 2021-05-25
AsyncAioPikaManager connect


wangjianweiwei:
https://github.com/miguelgrinberg/python-socketio/blob/2f0d8bbd8c4de43fe26e0f2edcd05aef3c8c71f9/socketio/asyncio_aiopika_manager.py#L68-L76

A connection is built every time it is publish can this place be improved?

miguelgrinberg:
Of course. This code was not written by useful, it is from a contributor. If you want to make improvements to it I'll be happy to merge them.

wangjianweiwei:
Thanks. I'm glad to contribute the code.

PR #1142, opened 2023-02-25
Use a persistent ampq connection when publishing

sillyfrog:
The current `AsyncAioPikaManager` implementation creates a new connection for _every_ publish message, which is very expensive in the AMPQ protocol.

This addresses that, and (hopefully) the concerns raised in #692.

I'm not an advanced (nor intermediate) user of `asyncio`, so any feedback or advice is very welcome.

Thanks for the great library!

Fixes:  #691

### DIFF
--- a/src/socketio/asyncio_aiopika_manager.py
+++ b/src/socketio/asyncio_aiopika_manager.py
@@ -44,6 +44,8 @@ class AsyncAioPikaManager(AsyncPubSubManager):  # pragma: no cover
                                '(Run "pip install aio_pika" in your '
                                'virtualenv).')
         self.url = url
+        self._lock = asyncio.Lock()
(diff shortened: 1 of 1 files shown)


sillyfrog:
Some how I missed #935 - Closing

sillyfrog:
Re-opening as I think this will be more robust vs #935 because of the lock around the connection creation

miguelgrinberg:
Thanks. This wasn't as robust as I would have hoped, and also you've introduced some code duplication. I have now implemented this optimization along with some code cleanup and corresponding changes in the Kombu code. Please try out the main branch and let me know if there are any remaining issues.

sillyfrog:
Testing the `async` build now, and it's looking good. Thanks!

That commit will also close #935 .

PR #1163, opened 2023-03-27
fix docstring typo: client/server mixup

Sasja:


Sasja:
sure, here you go, i overlooked it

miguelgrinberg:
Thanks!

PR #1230, opened 2023-08-30
Fix a typo in the documentation

spezifanta:
The code examples for passing a custom `http_session` to the client contain an incorrect reference. This commit corrects the reference to `requests.Session`.


miguelgrinberg:
Thanks!

PR #1284, opened 2023-12-27
fix and format by ruff

YoshitakaNaraoka:
As the title says


miguelgrinberg:
I can't accept this PR, which contributes nothing of value to the project.

PR #1288, opened 2023-12-29
Namespace catchall

### LINKED ISSUE #1285: Wildcard / catch-all namespace for event handlers
**Is your feature request related to a problem? Please describe.**
It's possible to accept connections on any namespace by specifying `namespaces='*'` in either Server init, but you can't actually handle arbitrary namespaces: unless a namespace is explicitly defined by an event handler, any client packets using it will get silently dropped.

**Describe the solution you'd like**
Allow specifying an event handler with `namespace='*'` to be a catch-all handler in the absence of an explicit namespace match.  In that case, the handler should also be given an explicit `namespace` parameter with the namespace from the client packet.

**Additional context**
Such a thing is already implemented for catch-all event handlers in the server's `_trigger_event` (e.g. https://github.com/miguelgrinberg/python-socketio/blob/6967eca5507abad23a1311594228168029455dd8/src/socketio/server.py#L610-L611), so similar logic could probably be used at the namespace level as well.

This is could perhaps allow a high level approach to #1166 with less library overhead since there'd be no regex comparison (though it could nonetheless be done by a user if so desired).

mooomooo:
Closes #1285 

Note: Existing tests all pass, ~~but no additional tests have been added validate wildcard functionality.~~
Namespace catchall tests added and also passed

### DIFF
--- a/src/socketio/async_server.py
+++ b/src/socketio/async_server.py
@@ -615,44 +615,9 @@ class AsyncServer(base_server.BaseServer):
         await self.manager.trigger_callback(sid, id, data)
 
     async def _trigger_event(self, event, namespace, *args):
-        """Invoke an application event handler.
-
-        Resolution priority:
-        - self.handlers[namespace][event]
-        - self.handlers[namespace]["*"]
-        - self.handlers["*"][event]
-        - self.handlers["*"]["*"]
-        - self.namespace_handlers[namespace]
-        - self.namespace_handlers["*"]
-        - self.not_handled
-        """
-        handler = None
+        """Invoke an application event handler."""
         # first see if we have an explicit handler for the event
-        if namespace in self.handlers and \
-                    event in self.handlers[namespace]:
-            handler = self.handlers[namespace][event]
-        elif namespace in self.handlers and \
-                    event not in self.reserved_events and \
-                    '*' in self.handlers[namespace]:
-            handler = self.handlers[namespace]['*']
-            args = (event, *args)
-        elif '*' in self.handlers and \
-                    event in self.handlers['*']:
-            handler = self.handlers['*'][event]
-            args = (namespace, *args)
-        elif '*' in self.handlers and \
-                    event not in self.reserved_events and \
-                    '*' in self.handlers['*']:
-            handler = self.handlers['*']['*']
-            args = (event, namespace, *args)
-        # or else, forward the event to a namepsace handler if one exists
-        elif namespace in self.namespace_handlers:  # pragma: no branch
-            return await self.namespace_handlers[namespace].trigger_event(
-                event, *args)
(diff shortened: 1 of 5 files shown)


mooomooo:
Added more tests, which should fix the codecov.  I don't know why the lint is failing -- I don't have any experience with tox and no error message is given.  @miguelgrinberg can you take a look?

Also, with a catchall namespace handler, you still need to explicitly set `namespaces` in the server initializer, which makes sense to me (so you can set a list of allowable namespaces to reject others while still taking advantage of the `'*'` functionality).  However, I don't think I had to do that for the event handlers?  That is, it didn't reject `s.manager.connect('123', '/bar')` even though I didn't set any handlers for `namespace = '/bar'`, and I didn't specify `namespaces` in the server initializer.  Is this intentional or a bug that needs to be tracked down somehow else?

miguelgrinberg:
@mooomooo the flake8 errors are these:

```
  src/socketio/base_server.py:210:21: E127 continuation line over-indented for visual indent
  src/socketio/base_server.py:213:21: E127 continuation line over-indented for visual indent
  src/socketio/base_server.py:214:21: E127 continuation line over-indented for visual indent
  src/socketio/base_server.py:218:21: E127 continuation line over-indented for visual indent
  src/socketio/base_server.py:222:21: E127 continuation line over-indented for visual indent
  src/socketio/base_server.py:223:21: E127 continuation line over-indented for visual indent
```

mooomooo:
Got it, makes sense.  I think this should be good for review now!

miguelgrinberg:
Looks good, nothing more to comment on. I need to test this out a little bit, write the documentation and then I'll merge it, hopefully within the next week or so. Thanks!

miguelgrinberg:
Merged! I copied your server implementation to the client and also added some documentation. Thanks!

kdNew23:
Thank you for working on this enhancement. Has this code been released yet? 

https://github.com/miguelgrinberg/python-socketio/blob/801241378ec6cbbaef6490399220b00473f66d34/src/socketio/async_server.py#L617 doesn't seem to have the latest code changes from [Codecov](https://app.codecov.io/gh/miguelgrinberg/python-socketio/pull/1288?src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Miguel+Grinberg) 

I'm trying to get catch-all namespaces for event handler to work.

```
@sio.on('connection_event', namespace='*')
async def connection_event(namespace, sid, data):
```

Also, for connect and disconnect events, I could get catch-all namespaces to work by looping through list of namespaces and defining decorators.

miguelgrinberg:
@kdNew23 Yes, this is in the latest release.

PR #1316, opened 2024-03-24
fix: the logic for getting the handler

Champgy:
First, get the handler from the specified namespace. If the handler is None, then get the handler from the namespace *.
#1315


miguelgrinberg:
Thanks!

PR #1329, opened 2024-04-27
Fix bug with packet.id equivalent to zero; check if message id is not None instead of boolean casting

dantetemplar:
Just check the commit


miguelgrinberg:
Out of curiosity, which Socket.IO client uses 0 as a callback id?

dantetemplar:
> Out of curiosity, which Socket.IO client or server uses 0 as a callback id?

I came across a bug while working with the JS client and msgpack plugin ^^

PR #1330, opened 2024-04-27
Integrate msgpack.Timestamp with python datetime.datetime by default

dantetemplar:


miguelgrinberg:
What is this trying to address? Date and datetime objects are not serializable. You should not use them in Socket.IO payloads.

dantetemplar:
> What is this trying to address? Date and datetime objects are not serializable. You should not use them in Socket.IO payloads.

They are serializable effectively under msgpack transport, but some arguments needed to be passed.

miguelgrinberg:
That is also true of any data type, and also possible to configure with JSON. But I prefer to avoid custom serializations. You can still use your custom packet class by passing it in the `serializer` argument when you create your server and client.

PR #1373, opened 2024-08-12
Mod: Change logger verbose level when handling emit and receive events

nghianv19940:
Send and receive events happen a lot. I want to exclude those from other events; if anyone wants to see those, use debug level.


miguelgrinberg:
Aside from making sense or not, this is a breaking change, so I cannot accept it right now and would only be open to consider it around the time of a major release.

PR #1374, opened 2024-08-17
Enable emitting to single client in managers with to=...

rustworthy:
Addresses #1375

### DIFF
--- a/src/socketio/async_manager.py
+++ b/src/socketio/async_manager.py
@@ -11,12 +11,13 @@ class AsyncManager(BaseManager):
         return self.is_connected(sid, namespace)
 
     async def emit(self, event, data, namespace, room=None, skip_sid=None,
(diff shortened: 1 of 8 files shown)


miguelgrinberg:
Thanks!

The problem here is that the managers expect the addressee of the message to be given in the `room` argument, which is an alias of `to` in the main server and client classes. For consistency, the `to` argument needs to be added in all manager subclasses, not just the one you fixed.

rustworthy:
> Thanks!
> 
> The problem here is that the managers expect the addressee of the message to be given in the `room` argument, which is an alias of `to` in the main server and client classes. For consistency, the `to` argument needs to be added in all manager subclasses, not just the one you fixed.

My pleasure!

Please see if I have identified all of them.

miguelgrinberg:
Thanks!

PR #1407, opened 2024-11-19
fix AsyncClient::wait unexpected return after success reconnect

arkuzo:
AsyncClient::wait use sleep(1) call to wait to start reconnect task. Sometimes reconnect is faster then 1 second, and wait returns while connection to server is established.

Added one check to method to avoid this situation


miguelgrinberg:
Makes sense, thanks!

Could I ask you to apply a similar fix to the synchronous client?

Issue #1276, opened 2023-11-08
Request: type hints (e.g. for mypy)


jamesbraza:
**Is your feature request related to a problem? Please describe.**

I run `mypy` on my code, and it's throwing errors for not being typed:

```
a.py:19:1: error: Skipping analyzing "socketio": module is installed, but missing library stubs or py.typed marker  [import-untyped]
    from socketio import Client
    ^
```

**Describe the solution you'd like**

Type hints to be added to the repo, or alternately a type stub package to be installable.

**Describe alternatives you've considered**

Disabling mypy for `python-socketio` like the below `pyproject.toml` config:

```toml
[[tool.mypy.overrides]]
# Suppresses error messages about imports that cannot be resolved.
ignore_missing_imports = true
# Per-module configuration options
module = ["socketio.*"]
```

miguelgrinberg:
Unfortunately I do not have the time to maintain type hints on all my packages at this time. Once the typing solution in Python becomes more manageable I will consider doing it, but for now my suggestion is that you contribute a stub package to typeshed.

andy-takker:
@miguelgrinberg Hello, can I create PR with adding typehints for this project?

miguelgrinberg:
My answer hasn't changed. You are welcome to contribute type hints for these package to typeshed.

phi-friday:
It's not complete yet, but I have created a stub package for type hints.
https://github.com/phi-friday/python-socketio-stubs
https://pypi.org/project/python-socketio-stubs/

PR #1409, opened 2024-11-25
Type hints for event handler

dokterbob:
Prevents type errors in library use.

Not 100% sure that the module-level definition of handlerparams/returns will not cause problems when different handler params are defined (e.g. handlers taking different parameters) - so feedback is 100% welcome.

Related: #1276


dokterbob:
@miguelgrinberg Sorry, I misread your last remark on related issue. I'll keep this for reference but will submit stubs to typeshed instead.

PR #1421, opened 2024-12-12
server.py: teeny docstring typo fix

masenf:
Noticed while perusing the documentation, so submitting a micropatch.


miguelgrinberg:
Thanks!

PR #1423, opened 2024-12-20
Add the reason argument for async_admin when socket disconnect

Yixiu1121:
Hi,

I am using python-socketio with an Admin UI and noticed that the disconnect reason is not being displayed. I have made some modifications to address this issue. Could you please review my pull request?

Thank you for your time and consideration!

### DIFF
--- a/src/socketio/async_server.py
+++ b/src/socketio/async_server.py
@@ -585,7 +585,7 @@ class AsyncServer(base_server.BaseServer):
         self.manager.pre_disconnect(sid, namespace=namespace)
         await self._trigger_event('disconnect', namespace, sid,
                                   reason or self.reason.CLIENT_DISCONNECT)
-        await self.manager.disconnect(sid, namespace, reason=reason or self.reason.CLIENT_DISCONNECT, ignore_queue=True)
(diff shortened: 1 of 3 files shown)


miguelgrinberg:
Good catch, I forgot that the Admin UI can show the reason. Thank you so much, I'll review your change and merge over the next few days.

src/socketio/async_server.py
```diff
@@ -585,7 +585,7 @@ async def _handle_disconnect(self, eio_sid, namespace, reason=None):
         self.manager.pre_disconnect(sid, namespace=namespace)
         await self._trigger_event('disconnect', namespace, sid,
                                   reason or self.reason.CLIENT_DISCONNECT)
-        await self.manager.disconnect(sid, namespace, ignore_queue=True)
+        await self.manager.disconnect(sid, namespace, reason=reason or self.reason.CLIENT_DISCONNECT, ignore_queue=True)
```
miguelgrinberg, line 588: What is this for?

miguelgrinberg:
@Yixiu1121 Thanks for contributing the fix. Unfortunately the solution needed to be more complicated than this, so I ended up implementing this fix separately.

PR #1472, opened 2025-07-03
V4.6.1.post1

sharanchetan:
We plan to use this library (4.6.x) with a new version of python i.e. 3.11+. There has been breaking changes in Asyncio version which causes the known issue of "Passing coroutines is forbidden, use tasks explicitly.".

You have already provided the fix for this issue in main branch which corresponds to later version of socket-io protocol ( protocol version 5 & library version 5.x.x).

We request you to include this change & release a version in pypi.org

### DIFF
--- a/socketio/asyncio_manager.py
+++ b/socketio/asyncio_manager.py
@@ -26,8 +26,8 @@ class AsyncManager(BaseManager):
                     id = self._generate_ack_id(sid, namespace, callback)
                 else:
                     id = None
-                tasks.append(self.server._emit_internal(sid, event, data,
-                                                        namespace, id))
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
The version that you are using is almost 5 years old. If you need such an old version then you are welcome to fork the library and create and maintain your own patched version. Unfortunately I do not have the resources to support old releases myself.

I honestly think the best would be for you to upgrade your other dependencies that are stuck with such an old version of the Socket.IO protocol so that you can upgrade to more recent versions of this package, because in 5 years there has been a lot of fixes and improvements that you don't have when using the 4.x versions.

PR #1476, opened 2025-07-22
Bugfix: self.channel not initialised

Eugnee:
_redis_connect expects self.channel to be initialised, otherwise getting exception "object has no attribute 'channel'" at line 70


miguelgrinberg:
Thanks!

PR #1479, opened 2025-07-28
redis manager, verbose error logging

Eugnee:
Sometimes it's very useful to see exact error from redis in logs


Eugnee:
I usually prefer full stack trace, but you probably right. 
In this case, I guess something like str(exc) should be fine.

Eugnee:
@miguelgrinberg Hi, how are you?
It's okay now, or you see it differently?

miguelgrinberg:
LGTM. Thanks!

Issue #1428, opened 2024-12-27
Support for valkey-py based client manager


Ohnoimded:
## Problem

Since Redis has been forked to [Valkey](https://github.com/valkey-io/valkey) and many projects are likely to adopt Valkey over Redis, using [RedisManager/AsyncRedisManager](https://python-socketio.readthedocs.io/en/latest/server.html#using-a-message-queue) no longer works with Valkey URLs (e.g., `valkey://...`) and [valkey-py](https://github.com/valkey-io/valkey-py), as it only checks for Redis URLs (e.g., `redis://...`).

## Solution

I created a custom implementation called `ValkeyManager/AsyncValkeyManager` inherited from `PubSubManager/AsyncPubSubManager`, using [valkey-py](https://github.com/valkey-io/valkey-py) instead of [redis-py](https://github.com/redis/redis-py).

## Why This Matters

Due to Redis' licensing change, many projects are transitioning to Valkey. Adding support for Valkey URLs would help users migrate seamlessly and avoid the need for custom solutions.

## Request

Would it be possible to integrate support for Valkey URLs in the official implementation? This would make it easier for projects to adopt Valkey.

miguelgrinberg:
Instead of duplicating the existing classes that work with Reids, I would prefer to extend them to support either the Redis or the Valkey clients. This would avoid code duplication, at least while the two projects mantain similar APIs.

Ohnoimded:
Sounds like a good idea. Can I make a PR for this?

miguelgrinberg:
@Ohnoimded Sure, go ahead and submit a PR once done.

PR #1488, opened 2025-08-21
feat: support valkey

phi-friday:
closes: #1428 

It has been half a year since the issue was raised, but I noticed that no related PR has been created, so I gave it a try.
I believe I made minimal changes, and I expect this will remain valid unless the APIs of Valkey and Redis change drastically.

### DIFF
--- a/src/socketio/async_redis_manager.py
+++ b/src/socketio/async_redis_manager.py
@@ -20,8 +20,6 @@ except ImportError:  # pragma: no cover
     valkey = None
     ValkeyError = None
 
-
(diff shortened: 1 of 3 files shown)


phi-friday:
@miguelgrinberg 
If there is any code that needs to be modified or improved, could you please let me know?

miguelgrinberg:
@phi-friday I'll have a look at this as soon as I can. The main thing that is missing is to restore full coverage, but I can take a look at this myself. Thanks!

PR #1499, opened 2025-09-28
SimpleClient: type mismatch for `SimpleClient.transport()`

jthistle:
Normally, `SimpleClient.transport` has signature `() -> str`. However, if `SimpleClient.connect()` is not called `SimpleClient.transport` is simply a `str` (not a function).

I came across this as this was confusing the Pylance type checker.

My proposed fix ensures reasonable behaviour by always making `SimpleClient.transport` a function. It shouldn't break anything important since `SimpleClient.transport` is meaningless before `connect` is called anyway. However, this is not a strictly backwards-compatible change.

### DIFF
--- a/src/socketio/simple_client.py
+++ b/src/socketio/simple_client.py
@@ -103,7 +103,7 @@ class SimpleClient:
         The transport is returned as a string and can be one of ``polling``
         and ``websocket``.
         """
-        return self.client.transport if self.client else ''
+        return self.client.transport if self.client else lambda: ''
 
     def emit(self, event, data=None):
         """Emit an event to the server.


jthistle:
Ok, cool. But won't doing that break backwards compatability for user code in a worse way, since now all calls to `SimpleClient.transport()` will raise exceptions? Why not just update the docstring instead?

Also, I'm not sure the docstring was necessarily incorrect as `SimpleClient.transport()` _did_ return a string, just as it stated.

miguelgrinberg:
This issue cannot be fixed in a backwards compatible manner. As I said above, the return value was not in agreement with the documented way this property works, and if backwards compatibility is going to be broken regardless of the fix, I rather fix the bug propertly instead of perpetuating an old mistake that I've made.

PR #1501, opened 2025-09-29
Fix: SimpleClient.call does not raise TimeoutError on timeout

jthistle:
Heya, back again. Pretty sure this one _is_ a bug. If the underlying `call` times out, it leads to an infinite loop since the `TimeoutError` is caught in the `except SocketIOError` block and ignored. 

I would think that we want to reraise that `TimeoutError`. (And maybe all other `SocketIOError`s too?)


miguelgrinberg:
Okay, sorry, I see the problem now. So what this PR is missing is a) similar fix on the async side, and b) unit tests. Is this something you can do?

PR #1502, opened 2025-09-30
Replace pickle with json

miguelgrinberg:
It has been reported (credit: Ali Raza, BlueRock) that the use of the `pickle` package from the Python standard library to encode payloads that are passed between Socket.IO processes over a message queue (in a multi-server and/or external process configuration) can be exploited to create a remote code execution.

For this exploit to be viable, the attacker needs to connect to the message queue directly, and push a malicious payload to it. Given that the message queue is an internal component of the server deployment that is not accessible from the public internet, the only reason for concern would be if the queue is configured to listen publicly by mistake.

To confirm that you have a secure deployment, make sure that:

1. Your message queue is only listening on an internal network interface (i.e localhost) or VPC. Never deploy the message queue for Socket.IO to the public internet. 
2. For added protection, configure your production message queue with authentication and encryption. Please consult the documentation of your message queue for instructions. The message queue credentials can then be added to the connection URL passed to python-socketio in the configuration. And of course, avoid committing a URL with credentials to source control.

I have added a section to the documentation on [secure deployment of the message queue](https://python-socketio.readthedocs.io/en/latest/server.html#deploying-the-message-queue-for-production) with the above recommendations.

In addition to the above, and considering that `pickle` does not provide any benefits over other serialization formats, I have decided to remove the use of pickle completely, so that any possibility of an attack, though unlikely, is eliminated. Version 5.10.0 uses JSON as the only communication format for message queue payloads (in previous versions both JSON and Pickle were supported, but Pickle was used by default).

The only side effect of this change is that a mix deployment that includes servers from before and after this change may fail to communicate, because the older versions will attempt to push Pickle packets to the queue and the newer servers do not understand that format anymore. If all servers are upgraded to 5.10 or newer there shouldn't be any problems.

### DIFF
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -1096,17 +1096,17 @@ For a production deployment there are a few recommendations to keep your
 application secure.
 
 First of all, the message queue should never be listening on a public network
-interface, to ensure that external clients never connect to it. The use of a
-private network (VPC), where the communication between servers can happen
-privately is highly recommended.
+interface, to ensure that external clients never connect to it. For a single
+node deployment, the queue should only listen on `localhost`. For a multi-node
+system the use of a private network (VPC), where the communication between
+servers can happen privately is highly recommended.
 
-In addition, all message queues support authentication and encryption.
-Authentication ensures that only the Socket.IO servers and related processes
-have access, while encryption prevents data to be collected by a third-party
-listening on the network.
-
-Access credentials can be included in the connection URLs that are passed to the
-client managers.
+In addition, all message queues support authentication and encryption, which
+can strenthen the security of the deployment. Authentication ensures that only
+the Socket.IO servers and related processes have access, while encryption
+prevents data from being collected by a third-party that is listening on the
+network. Access credentials can be included in the connection URLs that are
+passed to the client managers.
 
 Horizontal Scaling
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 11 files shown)


miguelgrinberg:
@judilsteve Thanks. These are problems that started after I dropped the use of Pickle, which handled all these serialization issues transparently. I'll make sure `bytesstrings` are accepted as binary, and will also look at the server issue.

miguelgrinberg:
@judilsteve I do not see the crash you reported above. Can you share the code that you are using that causes this crash on line 37 of async_manager.py? Thanks.

shmcgough:
I've also run into an issue since the switch to json was made. I'm using `AsyncRedisManager` and setting a tuple type as the data value no longer turns them into individual arguments for the (javascript) client, instead the first argument is an array of the arguments and the other arguments are undefined.

I can manually change the package code to go from pickle to json and see that it works one way and not the other.

miguelgrinberg:
@shmcgough Would you be able to install the main branch of this repo and test your application with it? I think the support for multiple arguments via tuples is now restored. I'll cut a release in a few days, so let me know if you experience any remaining issues. Thanks!

shmcgough:
@miguelgrinberg I was able to test the main branch and can confirm it works! Thanks again.

miguelgrinberg:
@shmcgough v5.15.1 is now available with this fix.

Issue #487, opened 2020-05-20
ConnectionRefusedError not working


databasedav:
This is on Python 3.8.2 with
```
python-engineio==3.12.1
python-socketio==4.5.1
uvicorn==0.11.5
```
and you can reproduce with
```python
# server.py
import socketio
from socketio.exceptions import ConnectionRefusedError

sock = socketio.ASGIApp(
    socketio.AsyncServer(async_mode="asgi"),
)

@sock.engineio_server.event
async def connect(sid, environ):
    # raise Exception()
    raise ConnectionRefusedError()
```
```python
# client.py
import asyncio
import socketio

async def test():
    c = socketio.AsyncClient()
    await c.connect('ws://localhost:8002')
    await asyncio.sleep(1)
    await c.disconnect()
    await asyncio.sleep(1)

asyncio.run(test())
```
and running `uvicorn --host 0.0.0.0 --port 8002 server:sock` and `python client.py`

raising an `Exception` produces the expected behavior.

miguelgrinberg:
Sorry, but I can't reproduce. I get the expected `ConnectionError` in the client using exact copies of your server and client:

```
socketio.exceptions.ConnectionError: Connection refused by the server
```

databasedav:
Hmm not sure why reproduction isn't working. On my end, downgrading to 4.3.1 solves the issue so it seems to be something introduced in 4.4.0. I'm on Ubuntu 20.04 but not sure if that is relevant. I'll investigate further.

miguelgrinberg:
Sorry, I gave your code another run today and I do see the problem. I must have put your server on the standard 8000 port yesterday, instead of the custom 8002 that you are using.

ysde:
Hi @miguelgrinberg 

I raise ConnectionRefusedError('authentication failed') in `on_connect` event.

But the client only got `ConnectionError('Connection error')` when I print the exception.

The goal is to get the `authentication failed` message to tell the client. Just like the instruction in the documentation.

```
Sometimes it is useful to pass data back to the client being rejected. In that case instead of returning False socketio.exceptions.ConnectionRefusedError can be raised, and all of its arguments will be sent to the client with the rejection message:

@sio.event
def connect(sid, environ):
    raise ConnectionRefusedError('authentication failed')
```
======================================================================
Here is the code

```
try:
    sio.connect('http://xxxxxxxxxxxxx')
except ConnectionError as ex:
    print(repr(ex))
```

[...]

PR #1512, opened 2025-10-26
docs: fix example in server.rst with more explicit exception handling

voiduin:
[@miguelgrinberg](https://github.com/miguelgrinberg), thank you for this project - it's very useful!

I started using python-socketio in my project but encountered a bug that took me considerable time to debug.

My client wasn't receiving the `connect_error` event, and I tried changing server/client versions, reviewing my code multiple times, and other troubleshooting steps.

Thanks to [this comment](https://github.com/miguelgrinberg/python-socketio/issues/487#issuecomment-2147293485), I was able to identify the issue.

image

I noticed that even in your own test files, you don't use this shortened version anywhere:
```
            side_effect=exceptions.ConnectionRefusedError('fail_reason')
```

Reference: https://github.com/miguelgrinberg/python-socketio/blob/f8a1d435b23df74e720cb1c384955ff5dd1c442c/tests/async/test_server.py#L474C13-L474C73

**Suggestion:** Please update the documentation example to use the full exception name instead of the shortened version.

Using the full module path (e.g., `socketio.exceptions.ConnectionRefusedError`) would make the examples consistent with your own codebase and much clearer for newcomers to understand.Retry

### DIFF
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -248,7 +248,7 @@ message::
 
     @sio.event
     def connect(sid, environ, auth):
-        raise ConnectionRefusedError('authentication failed')
+        raise socketio.exceptions.ConnectionRefusedError('authentication failed')
 
 The disconnect handler receives the ``sid`` assigned to the client and a
 ``reason``, which provides the cause of the disconnection::


miguelgrinberg:
I'm sorry about this, I did not pick up the inconsistency between the two exception classes when it was mentioned before. I think instead of changing the example we can do better and handle both classes. Is this something you'd like to attempt yourself, or should I do it myself?

PR #1516, opened 2025-10-30
fix recreate binding on failure

### LINKED ISSUE #1469: recreate bindings on failures
## Issue

Binding to an exchange should happen on every failure in the listening loop.
I've seen several times (not sure what exactly triggers it), that if rabbitmq goes down and comes up again, the bindings are lost.
If the "while True listen loop" then still runs it will never listen again to the queue again.
This is due to the fact that the rabbitmq exchange binding / channel creatoin happens before the while loop.

This ends up in something similar to a split brain.
The machine is able to send events to a room but never receives any.

Why not, on specific errors  or every time, recreate the binding to the exchange?

### LINKED ISSUE #1513: AsyncAioPikaManager passes a str to aio_pika.Message (expects bytes)
**Describe the bug**
When using an AsyncioServer instance with AsyncAioPikaManager as client_manager, on an emit call the client_manager tries to construct an aio-pika Message object using the type `str` for the body argument. The aio_pika.Message constructor expects a `bytes` body. Passing a `str` causes aio-pika to attempt a type-conversion using `bytes(body)`. Due to the missing encoding a TypeError is raised.
AsyncAioPika should pass the correct type `bytes` to the Message Constructor.

**To Reproduce**
Using python 3.13.5
Using python-socketio==5.14.2
Using aio-pika==9.5.7
1. Create a socketio.AsyncServer instance using the socketio.AsyncAioPikaManager

```
sio = socketio.AsyncServer(
    client_manager=socketio.AsyncAioPikaManager(settings.SETTINGS_SCHEMA.CELERY.url),
    async_mode="asgi",
)
```

2. Call the emit method on this instance using any string as event argument and None as data.

```
await sio.emit("dummy_event")
```

Minimal example
```
import asyncio
import socketio


async def main():
    server = socketio.AsyncServer(
        client_manager=socketio.AsyncAioPikaManager("amqp://localhost//"),
        async_mode="asgi",
        logger=True,
        engineio_logger=True,
    )
    await server.emit("dummy_event", data=None)


if __name__ == "__main__":
    asyncio.run(main())
```

**Expected behavior**
The Server to emit this event without raising a TypeError.

**Logs**
```
Server initialized for asgi.
emitting event "dummy_event" to all [/]
 ...
  File "/app/sample_sio_server.py", line 12, in main
    await server.emit("dummy_event", data=None)
  File "/app/.venv/lib/python3.13/site-packages/socketio/async_server.py", line 178, in emit
    await self.manager.emit(event, data, namespace, room=room,
                            skip_sid=skip_sid, callback=callback,
                            ignore_queue=ignore_queue)
  File "/app/.venv/lib/python3.13/site-packages/socketio/async_pubsub_manager.py", line 78, in emit
    await self._publish(message)  # notify other hosts
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.13/site-packages/socketio/async_aiopika_manager.py", line 84, in _publish
    aio_pika.Message(
    ~~~~~~~~~~~~~~~~^
        body=json.dumps(data),
        ^^^^^^^^^^^^^^^^^^^^^^
        delivery_mode=aio_pika.DeliveryMode.PERSISTENT
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ), routing_key='*',
    ^
  File "/app/.venv/lib/python3.13/site-packages/aio_pika/message.py", line 185, in __init__
    self.body = body if isinstance(body, bytes) else bytes(body)
                                                     ~~~~~^^^^^^
TypeError: string argument without an encoding
```

**Possible workarounds**

Either:
Downgrade to python-socketio==5.13.0

Or:
Create a CustomClientManager class, overwriting the `_publish` method with the Message Constructor encoding the body.

```
class CustomClientManager(socketio.AsyncAioPikaManager):
   def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
   
    async def _publish(self, data):
        if self.publisher_connection is None:
            async with self._lock:
                if self.publisher_connection is None:
                    self.publisher_connection = await self._connection()
                    self.publisher_channel = await self._channel(
                        self.publisher_connection
                    )
                    self.publisher_exchange = await self._exchange(
                        self.publisher_channel
                    )
        retry = True
        while True:
            try:
                await self.publisher_exchange.publish(
                    aio_pika.Message(
                        body=json.dumps(data).encode(),  # the only line changed from the parent _publish method
                        delivery_mode=aio_pika.DeliveryMode.PERSISTENT
                    ), routing_key='*',
                )
                break
            except aio_pika.AMQPException:
                if retry:
                    self._get_logger().error('Cannot publish to rabbitmq... '
                                             'retrying')
                    retry = False
                else:
                    self._get_logger().error(
                        'Cannot publish to rabbitmq... giving up')
                    break
            except aio_pika.exceptions.ChannelInvalidStateError:
                # aio_pika raises this exception when the task is cancelled
                raise asyncio.CancelledError() 

```

codomposer:
# Fix: Recreate RabbitMQ bindings on connection failure

## Fixes

Closes #1469 
Closes #1513 

## Problem

When RabbitMQ goes down and comes back up, the Socket.IO servers enter a **split-brain state** where they can send events but never receive them. This occurs because:

1. Exchange/queue bindings are created **once** before the listening loop
2. When RabbitMQ restarts, these bindings become invalid
3. The retry logic reconnects but **reuses the stale queue/exchange references**
4. Messages can be published (new connections work) but never consumed (old bindings are dead)

This results in servers that appear healthy but silently fail to process incoming events.

## Root Cause

When RabbitMQ goes down and comes back up, Socket.IO servers enter a **split-brain state** where they can send events but never receive them.

## Solution

Move the exchange/queue binding creation inside the retry loop so they are recreated on every connection failure.

## Result

When RabbitMQ restarts, the exception triggers a retry that creates fresh connection, channel, exchange, and queue objects with valid bindings. Servers can now both send and receive events after recovery.

### DIFF
--- a/src/socketio/async_aiopika_manager.py
+++ b/src/socketio/async_aiopika_manager.py
@@ -101,26 +101,26 @@ class AsyncAioPikaManager(AsyncPubSubManager):  # pragma: no cover
                 raise asyncio.CancelledError()
 
     async def _listen(self):
-        async with (await self._connection()) as connection:
-            channel = await self._channel(connection)
-            await channel.set_qos(prefetch_count=1)
-            exchange = await self._exchange(channel)
-            queue = await self._queue(channel, exchange)
+        retry_sleep = 1
+        while True:
+            try:
+                async with (await self._connection()) as connection:
+                    channel = await self._channel(connection)
+                    await channel.set_qos(prefetch_count=1)
+                    exchange = await self._exchange(channel)
+                    queue = await self._queue(channel, exchange)
 
-            retry_sleep = 1
-            while True:
-                try:
                     async with queue.iterator() as queue_iter:
                         async for message in queue_iter:
                             async with message.process():
                                 yield message.body
                                 retry_sleep = 1
-                except aio_pika.AMQPException:
-                    self._get_logger().error(
-                        'Cannot receive from rabbitmq... '
-                        'retrying in {} secs'.format(retry_sleep))
-                    await asyncio.sleep(retry_sleep)
-                    retry_sleep = min(retry_sleep * 2, 60)
-                except aio_pika.exceptions.ChannelInvalidStateError:
-                    # aio_pika raises this exception when the task is cancelled
-                    raise asyncio.CancelledError()
+            except aio_pika.AMQPException:
+                self._get_logger().error(
+                    'Cannot receive from rabbitmq... '
+                    'retrying in {} secs'.format(retry_sleep))
+                await asyncio.sleep(retry_sleep)
+                retry_sleep = min(retry_sleep * 2, 60)
+            except aio_pika.exceptions.ChannelInvalidStateError:
+                # aio_pika raises this exception when the task is cancelled
+                raise asyncio.CancelledError()
--- a/src/socketio/kombu_manager.py
+++ b/src/socketio/kombu_manager.py
@@ -115,10 +115,10 @@ class KombuManager(PubSubManager):  # pragma: no cover
                     break
 
     def _listen(self):
-        reader_queue = self._queue()
         retry_sleep = 1
         while True:
             try:
+                reader_queue = self._queue()
                 with self._connection() as connection:
                     with connection.SimpleQueue(reader_queue) as queue:
                         while True:


julianhille:
when is this going to be released? this is a huge improvement for us apparently.

PR #1517, opened 2025-11-02
feat: support to msgpack ext type

phi-friday:
This change adds two new parameters, `dumps_default` and `ext_hook`, to enable support for msgpack ext type.
With these arguments, `MsgPackPacket` can encode and decode values using msgpack ext type.

### DIFF
--- a/src/socketio/msgpack_packet.py
+++ b/src/socketio/msgpack_packet.py
@@ -1,5 +1,8 @@
+import logging
 import msgpack
 from . import packet
+
+logger = logging.getLogger('socketio')
 
 
 class MsgPackPacket(packet.Packet):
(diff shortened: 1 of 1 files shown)


phi-friday:
> Ah — after writing the sample code I realized that because this class is being passed, it's not possible to provide arguments to its `__init__`. The places that use it will also need to be modified, or an alternative approach is required. Simply changing this class alone doesn't solve the problem.

I think if the user uses `functools.partial`, no modifications should be necessary.

miguelgrinberg:
Partial works, but I don't see it as a good solution. If anything it would make every packet creation slower.

I have to think about this. I think passing these new arguments in the constructor of the packet is not the most usable solution. These have to be configured elsewhere and globally.

miguelgrinberg:
I've been thinking about this, and I'd like to have a simpler interface. For example:

```python
serializer=MsgPackPacket  # use with default options
serializer=MsgPackPacket.configure(default=my_default, ext_hook=my_ext_hook)  # custom configuration
```

This makes the configuration explicit, and visible to IDEs and type checkers, versus the `serializer_args` approach which is going to be harder for people to figure out, and not possible to type check for those interested in that.

Unfortunately implementing my the above pattern is not trivial, as it requires generating new classes on the fly. I'll see if I can modify this PR to do it over the next few days.

PR #1532, opened 2025-11-12
AsyncRedisManager: set safe pub/sub defaults (fix Valkey socket_timeout spam)

yokha:
> **Context**
> Migrating from Redis to Valkey (`python-valkey==6.1.1`) using:
>
> ```py
> AsyncRedisManager(url=f"valkey://{host}")
> ```
>
> triggers repeated:
>
> ```
> ERROR - socketio.server - Cannot receive from redis... retrying in 1 secs
> ```
>
> **Root cause**
> Valkey’s default `socket_timeout` is `5` (seconds), while Redis’ default is `None`. A 5s timeout forces the blocking pub/sub read to raise periodically; AsyncRedisManager treats that as an error and reconnects, causing log spam.
>
> **Change**
> Pin sane pub/sub defaults inside `AsyncRedisManager._redis_connect()` and merge with `redis_options` (user overrides last):
>
> * `socket_timeout = None` (block indefinitely)
> * `health_check_interval = 0` (no periodic PINGs on the read socket)
> * `decode_responses = False` (channel bytes)
> * `retry_on_timeout = False` (let AsyncRedisManager’s listen loop handle reconnect/backoff)
>
> **Why**
> Avoid relying on backend client defaults that differ (and may change over time). This makes pub/sub behavior stable across Redis and Valkey while preserving user configurability.
>
> **Backward compatibility**
>
> * Redis: effective behavior unchanged.
> * Valkey: eliminates timeout/reconnect spam.
> * Users can override any default through `redis_options`.
>
> **Tests**
>
> * Cover kwarg merging for `redis://`, `valkey://`, and `*+sentinel` URLs.
>
> No style-only changes; scope limited to pub/sub defaults + tests.

### DIFF
--- a/src/socketio/async_redis_manager.py
+++ b/src/socketio/async_redis_manager.py
@@ -95,16 +95,23 @@ class AsyncRedisManager(AsyncPubSubManager):
     def _redis_connect(self):
         module, _ = self._get_redis_module_and_error()
         parsed_url = urlparse(self.redis_url)
+        pubsub_defaults = {
+            "decode_responses": False,   # keep channels as bytes
+            "socket_timeout": None,      # block indefinitely on pub/sub
+            "health_check_interval": 0,  # no periodic PINGs on read socket
+            "retry_on_timeout": False,   # Pub/Sub loop handles reconnects
+        }
+        kwargs = {**pubsub_defaults, **(self.redis_options or {})}
+
         if parsed_url.scheme in {"redis+sentinel", "valkey+sentinel"}:
             sentinels, service_name, connection_kwargs = \
                 parse_redis_sentinel_url(self.redis_url)
-            kwargs = self.redis_options
-            kwargs.update(connection_kwargs)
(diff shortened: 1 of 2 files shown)


miguelgrinberg:
@yokha I'm looking at the Valkey defaults in the code, and they are the same as in Redis.

Valkey (https://github.com/valkey-io/valkey-py/blob/main/valkey/client.py#L172):

```python
socket_timeout=None,
health_check_interval=0,
decode_responses=False,
retry_on_timeout=False,
```

Redis (https://github.com/redis/redis-py/blob/master/redis/client.py#L202):

```python
socket_timeout: Optional[float] = None,
health_check_interval: int = 0,
decode_responses: bool = False,
retry_on_timeout: bool = False,
```

miguelgrinberg:
> Just to check before submitting anything new: do you think it’s worth adding a small fix in the listen() loop to handle str messages gracefully, or is it the intended behavior to silently skip non-byte messages?

Please write an issue describing the problem and reproduction instructions and we'll go from there.

PR #1533, opened 2025-11-12
AsyncRedisManager: reconnect loop owns subscribe/unsubscribe lifecycle (fix intermittent pub/sub stalls)

yokha:
Summary
-------
This PR confines pub/sub subscribe/unsubscribe to the reconnect loop, so the
listener always (re)subscribes after a backend error and cleanly unsubscribes
when the async generator closes. This fixes cases where a first backend
TimeoutError (e.g., Valkey with non-None defaults) left the subscription
lifecycle mismatched and caused repeated “retrying in 1s” logs.

What changed
------------
- `_redis_listen_with_retries` now:
  - Performs the initial `subscribe` and all re-subscriptions after reconnects.
  - Catches backend errors (Redis/Valkey), `OSError`, `TimeoutError`,
    backs off, and reconnects.
  - Ensures `unsubscribe` runs in a `finally` so early closes don’t leak
    subscriptions.
- `_listen` no longer subscribes/unsubscribes itself; it delegates that to the
  reconnect loop and only filters messages.

Why
---
Previously, the first `subscribe` happened in `_listen()` while reconnects and
re-subscriptions happened in `_redis_listen_with_retries()`. On some backends
(Valkey with a finite socket timeout) the first timeout could leave the system
in a state where the reconnect loop listened without a fresh subscription,
spamming “retrying in 1s”.

Tests
-----
- Added lifecycle tests that fail with the old flow and pass now:
  - Reconnects after a simulated timeout and yields the next message.
  - Closing the async generator triggers an `unsubscribe`.

Scope
-----
No behavioral change for healthy connections; improves robustness on transient
errors across Redis/Valkey backends. No API changes.

### DIFF
--- a/src/socketio/async_redis_manager.py
+++ b/src/socketio/async_redis_manager.py
@@ -1,4 +1,5 @@
 import asyncio
+import contextlib
 from urllib.parse import urlparse
 
 try:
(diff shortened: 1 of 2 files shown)


yokha:
Hi Miguel,
Thanks for the review.
Let me clarify what the PR fixes, because it is not about “missing subscriptions”
but about **where subscribe-time backend failures are handled**.

### 1. The problem is not functional breakage, it’s I guess incorrect layering

Today:

* `subscribe()` is executed inside `_listen()` (no try/catch block)
* If it raises (TimeoutError / OSError / backend stall),
  the exception is **not caught there**
* It propagates into `AsyncPubSubManager._thread()`

This produces:

```
Unexpected Error in pubsub listening thread
```

which is a manager-layer error handling a **backend-layer** exception.

The system does recover — the manager thread restarts (continue the loop)—
but recovery happens at the *wrong* abstraction level.

### 2. Why this matters

Because of that layering mismatch:

**A) The first subscribe failure bypasses the retry/backoff logic entirely.**
It never goes through `_redis_listen_with_retries()`.

**B) The logs are misleading** (generic manager error instead of backend retry).

**C) Lifecycle is inconsistent**
(first subscribe happens in `_listen()`, reconnect subscribe in `_redis_listen_with_retries()`).

**D) Repeated subscribe failures put the pubsub manager in a tight crash loop.**

This is easy to reproduce:
[https://github.com/yokha/python-socketio-valkey-playground](https://github.com/yokha/python-socketio-valkey-playground)

Set `fail_subscribe_attempts = 10000` and you get:

* immediate crash in `_thread()`
* immediate restart
* no backoff
* very high churn
* no active subscriber during this period

This isn’t a fatal bug, but it is a **design issue**:
subscribe-time errors should not be surfaced at the manager thread level.

### 3. What the PR changes

The PR is not a workaround for one backend.
It simply restores **consistent lifecycle ownership**:

* `_redis_listen_with_retries()` now owns:

  * connect
  * subscribe
  * listen
  * backend errors
  * backoff
  * reconnect
  * unsubscribe (in a `finally:`)

* `_listen()` is reduced to message filtering only

* `_thread()` is no longer catching backend exceptions

This aligns the implementation with typical async reconnection patterns:
**the layer that owns the connection also owns subscribe/unsubscribe and failure handling.**

### 4. Summary

The current implementation works, but handles subscribe failures at the wrong layer.
The PR corrects the lifecycle flow so that:

* first subscribes and re-subscribes follow the same retry logic
* backend failures don’t bubble into `_thread()`
* backoff is applied consistently
* repeated failures do not put the pubsub manager in a tight crash loop

>Note: The PR does **not** magically recover dropped pubsub messages.
> It ensures failures are handled in the **correct** place with proper backoff.

Happy to adjust any details, but the motivation is purely architectural:
**subscribe/unsubscribe should be part of the same lifecycle state machine as reconnects.**

miguelgrinberg:
@yokha I understand this well enough now. It would be much easier if you spoke by yourself instead of sending your LLM to talk to me.

Your change is not good, I'm going to close this PR. I'll open an issue to keep track of the actual issue, and in due time I'll address this properly. Thanks.

